PR #11, opened 2023-12-14
Streaming utf-8

### LINKED ISSUE #9: `sendMessageStream` with Chinese prompt returns malfunctioning chunks
## Expected Behavior
`sendMessageStream` with Chinese prompt should returns right chunks.

## Actual Behavior
`sendMessageStream` with Chinese prompt returns malfunctioning chunks which contains `���`.

## Steps to Reproduce the Problem

1.  
    ```js
    const prompt = '创作一首古诗';
    const chat = await model.startChat(history);
    const result = await chat.sendMessageStream(prompt);
    let text = ''
    for await (const chunk of result.stream) {
      text += chunk.text();
    }
    console.log(text)
    ```
1. get result like this:
   ```
    **寒夜**
    
    月寒山冷夜漫长，
    秋声���瑟送寒凉。
    银辉霜雪遍地洒，
    红叶舞尽染霜霜。
    
    寒风凛冽凌人骨，
    万籁俱寂月霜寒。
    古刹钟声伴孤影，
    一曲悲歌满心酸。
    
    寒江衰柳起愁思，
    霜梅雪月亦难禁。
    孤篇残卷叹今昔，
    满目疮痍叹世事。
    
    寒������滚������来，
    哀鸿遍野亦不悔。
    纵然前路多险阻，
    岂能轻言半途弃？
   ```
1. text contains `���`

## Specifications

- Version: ^0.1.1
- Platform: macOS

alx13:
Fixes #9


packages/main/src/requests/stream-reader.ts
```diff
@@ -75,9 +75,27 @@ export function processStream(response: Response): GenerateContentStreamResult {
  * GenerateContentResponse in each iteration.
  */
 function readFromReader(
-  reader: ReadableStreamDefaultReader,
+  bodyReader: ReadableStreamDefaultReader,
 ): ReadableStream<GenerateContentResponse> {
   let currentText = "";
+  const reader = new ReadableStream<Uint8Array>({
```
hsubox76, line 81: Since `response.body` is a ReadableStream<Uint8Array> I think we can just pass `body` to this function and here we can do `body.pipeThrough(newTextDecoderStream()).getReader()`. But for the tests we'll have to rewrite mockResponse to really make `body` a ReadableStream instead of an empty object with a getReader() function

alx13: And it's not working in Node.js v21.2.0 - fetch body request have only one `getReader` method :(

packages/main/src/requests/stream-reader.ts
```diff
@@ -87,20 +105,22 @@ function readFromReader(
...
+          currentText += value;
           const match = currentText.match(responseLineRE);
+          // TODO: This needs to be refactored
           if (match) {
             let parsedResponse: GenerateContentResponse;
             try {
               parsedResponse = JSON.parse(match[1]);
+              controller.enqueue(parsedResponse);
+              currentText = currentText.slice(
+                "data: ".length + match[1].length + "\r\n".length,
```
alx13, line 117: @hsubox76 just in case, please check for off-by-one error

hsubox76: Checks out.

packages/main/src/requests/stream-reader.ts
```diff
@@ -110,6 +132,37 @@ function readFromReader(
   return stream;
 }
 
+function tryParse(currentText: string): {
+  leftText: string;
```
hsubox76, line 136: Can we call this "remainingText"?

hsubox76 (approved):

PR #15, opened 2023-12-15
Catch sendMessageStream() rejections

hsubox76:
Fixes https://github.com/google/generative-ai-js/issues/7

Errors in `sendMessageStream` caused by `generateContentStream` throwing should already be catchable as that promise is directly returned. Errors further downstream in the `this._sendPromise` chain (syntax errors while writing to history, errors while reading the stream) should be caught and logged with `console.error`. Added tests.


packages/main/src/methods/chat-session.ts
```diff
@@ -155,6 +165,16 @@ export class ChatSession {
             );
           }
         }
+      })
+      .catch((e) => {
+        // Errors in streamPromise are already catchable by the user as
```
lahirumaramba, line 170: Did you mean to say `Errors in this._sendPromise...` here?

lahirumaramba (approved):

packages/main/src/methods/chat-session.ts
hsubox76: No, streamPromise, the second step in the `this._sendPromise` chain. Everything after that isn't catchable by the user and is being handled and logged here.

PR #19, opened 2023-12-15
Obscure API key in error messages

hsubox76:
Create a RequestUrl class that has fine grained access to url parts and can selectively obscure them.

Fixes https://github.com/google/generative-ai-js/issues/8


lahirumaramba (approved):
LGTM! 🚀🚀

PR #20, opened 2023-12-15
Fix release workflow

hsubox76:
Specify registry, fix typo

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -43,13 +43,14 @@ jobs:
       - name: Create .npmrc
         run: |
           cat << EOF > "$HOME/.npmrc"
+            @google:registry=https://wombat-dressing-room.appspot.com
             //wombat-dressing-room.appspot.com/:_authToken=$NPM_TOKEN
           EOF
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -43,13 +43,14 @@ jobs:
       - name: Create .npmrc
         run: |
           cat << EOF > "$HOME/.npmrc"
+            @google:registry=https://wombat-dressing-room.appspot.com
```
lahirumaramba, line 46: I don't think we mention the registry in the Admin Node.js release workflow... is this a slightly different process?

lahirumaramba (approved):

.github/workflows/publish.yml
hsubox76: This repo is using the `changeset publish` command and not `npm publish` directly. Firebase Admin Node uses `npm publish` which has a flag where you can specify the registry: https://github.com/firebase/firebase-admin-node/blob/6bde095be8d8e98f5f48eaff4cf260c8b82109bf/.github/scripts/publish_package.sh#L23 but the `changeset` command doesn't have any such flag, so you have to put it into the `.npmrc`.

PR #21, opened 2023-12-15
Send API key in header instead of query param.

hsubox76:
Matches other platform SDKs, keeps API key out of url.


lahirumaramba (approved):
Oh this is actually a lot better. Sorry for not pointing this out in the previous PR
LGTM!

PR #31, opened 2023-12-19
Feature Addition: Introduce Request Timeout Configuration

WillianAgostini:
This pull request introduces a new feature that allows users to configure request timeouts and other properties within the RequestInit object. Now, developers can not only customize the timeout parameter but also leverage other relevant properties within the object, enhancing the overall flexibility and control over requests in the application. This enhancement ensures that users have the capability to fine-tune various aspects of requests, providing a more versatile and adaptable solution for diverse use cases.


packages/main/src/requests/request.ts
```diff
@@ -62,17 +62,21 @@ function getClientHeaders(): string {
...
   let response;
   try {
     response = await fetch(url.toString(), {
-      method: "POST",
-      headers: {
-        "Content-Type": "application/json",
-        "x-goog-api-client": getClientHeaders(),
-        "x-goog-api-key": url.apiKey,
+      ...requestInit,
+      ...{
```
hsubox76, line 71: Why are the default properties put into an object and re-spread?

packages/main/src/methods/count-tokens.ts
```diff
@@ -22,8 +22,13 @@ export async function countTokens(
   apiKey: string,
   model: string,
   params: CountTokensRequest,
+  requestInit?: RequestInit,
```
hsubox76, line 25: Can we name this `fetchOptions`? I know the TypeScript type is RequestInit but documentation for `fetch` usually calls this param `options` so users might not realize this is the same set of params usually passed to `fetch`.

hsubox76 (commented):
1) None of these methods are exposed directly to the user, they're all wrapped by methods in GenerativeModel (/packages/main/src/models/generative-model.ts), so the option would need to be added to all the methods there. Given that, there's a possibility to put these options on the GenerativeModel class itself so the user doesn't have to set them on every function call, but if we think users will likely want to set different options on every call, then we won't want to do that.

2) If this is going to apply to all requests, it needs to be added to ChatSession's sendMessage and sendMessageStream methods as well.

3) There's some housekeeping that needs to be done - run `yarn changeset` to generate a changeset and `yarn docs` to update reference docs - but that can wait until the other questions have been resolved.

packages/main/src/requests/request.ts
WillianAgostini: it was a mistake, commit 748ac95c6ee20c7b234632f81e74cefc09d0d039 solve this

packages/main/src/methods/count-tokens.ts
WillianAgostini: Yes, ebabefc53fe4123e75e78eb52fe4d808215a32a0

WillianAgostini:
I truly see that it is better suited for the GenerativeModel, as all requests for that model will follow the same behavior. So, the approach would be to add a new parameter to the getGenerativeModel method, called fetchOptions.

WillianAgostini:
> 1. None of these methods are exposed directly to the user, they're all wrapped by methods in GenerativeModel (/packages/main/src/models/generative-model.ts), so the option would need to be added to all the methods there. Given that, there's a possibility to put these options on the GenerativeModel class itself so the user doesn't have to set them on every function call, but if we think users will likely want to set different options on every call, then we won't want to do that.
> 2. If this is going to apply to all requests, it needs to be added to ChatSession's sendMessage and sendMessageStream methods as well.
> 3. There's some housekeeping that needs to be done - run `yarn changeset` to generate a changeset and `yarn docs` to update reference docs - but that can wait until the other questions have been resolved.

I truly see that it is better suited for the GenerativeModel, as all requests for that model will follow the same behavior. So, the approach would be to add a new parameter to the getGenerativeModel method, called fetchOptions.

mayneyao:
I think that OpenAI's SDK  offers a good example, allowing developers to pass parameters such as timeout, baseurl, and maxRetryTime when initializing the client. This makes it more flexible.

https://github.com/openai/openai-node/blob/b595cd953a704dba2aef4c6c3fa431f83f18ccf9/src/index.ts#L107

https://github.com/google/generative-ai-js/issues/25

WillianAgostini:
Hi @hsubox76 
Is it more advantageous to provide the fetchOptions attribute with its properties, or is it more efficient to add only the timeout property and implement it internally with the abortController?

hsubox76 (commented):
Thanks for doing this! Looks good at a glance. I need to take some time to review the main part of it and probably also consult with other platform teams (Python, the Vertex JS SDK, etc) to make sure we're all consistent. A couple of small questions in the meantime while I am working on getting that done.

packages/main/package.json
```diff
@@ -28,7 +28,7 @@
     "test:node:integration": "yarn build && TS_NODE_COMPILER_OPTIONS='{\"module\":\"commonjs\"}' mocha \"test-integration/node/**/*.test.ts\"",
     "lint": "eslint -c .eslintrc.js '**/*.ts' --ignore-path '../../.gitignore'",
     "api-report": "api-extractor run --local --verbose",
-    "docs": "yarn build && yarn api-documenter markdown -i ./temp -o ../../docs/reference/"
+    "docs": "yarn build && npx api-documenter markdown -i ./temp -o ../../docs/reference/"
```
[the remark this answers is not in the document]
WillianAgostini: it was an oversight...

.changeset/rotten-queens-cheat.md
```diff
@@ -0,0 +1,5 @@
+---
+"@google/generative-ai": patch
```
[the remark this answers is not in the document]
WillianAgostini: Ok

hsubox76:
Just wanted to update that we finally got approval to add the RequestOptions object and the final design agreed upon seems to match up to what you have implemented, so I think we should be good to merge. I'm going to look it over one more time to make sure we haven't missed anything and then will merge.

hsubox76 (approved):

tmkx:
can we add a new `signal` option? it will be more flexible, and meet the timeout requirement as well, `timeout` is also very useful as it's more convenient :)

```ts
const model = genAI.getGenerativeModel({ model: "gemini-pro" }, {
  signal: AbortSignal.timeout(10000),
});
```

https://github.com/google/generative-ai-js/blob/d8f22d2259c9e926dcc4ac955db896c235e0a3bc/packages/main/src/requests/request.ts#L103-L117

hsubox76:
Can you open a new issue titled "Feature Request: signal option" (or whatever description you think of) so we don't lose track of it, as this PR is closed.

Regarding this PR, version 0.2.0 is released and includes it.

PR #32, opened 2023-12-19
Function calling

alx13:
Initial revision for function calling

https://ai.google.dev/docs/function_calling


liranrevivo:
Will it be possible to use this feature of function calling along with the multi step (i.e. ChatSession) capabilities? e.g. to call the endpoint .startChat along with an API schema for the model to do function calling?

hsubox76 (commented):
Still looking through this, pending what the automatic function calling design is going to look like but FYI, sometime before this is merged, add a changeset (`yarn changeset`), should be a minor bump.

packages/main/src/requests/response-helpers.ts
```diff
@@ -54,6 +55,30 @@ export function addHelpers(
...
+      if (hadBadFinishReason(response.candidates[0])) {
+        throw new GoogleGenerativeAIResponseError<GenerateContentResponse>(
+          `${formatBlockErrorMessage(response)}`,
+          response,
+        );
+      }
+      return getFunctionCall(response);
+    } else if (response.promptFeedback) {
+      throw new GoogleGenerativeAIResponseError<GenerateContentResponse>(
+        `Text not available. ${formatBlockErrorMessage(response)}`,
```
hsubox76, line 76: Should probably change the message string to "Function call not available"?

packages/main/src/requests/response-helpers.ts
```diff
@@ -54,6 +55,30 @@ export function addHelpers(
     }
     return "";
   };
+  (response as EnhancedGenerateContentResponse).functionCall = () => {
```
hsubox76, line 58: The Vertex Python design for this convenience method is `functionCalls`, (internal link: https://docs.google.com/document/d/1Y2ncToFbzd1Ti1WJOXQHHAvFzCbNWzZYVleoDbcLrDc/edit?tab=t.0#heading=h.f0d2sodbkxt3) anticipating multiple function calls being returned in the future (see b/319098239).

But, if our philosophy is that these convenience methods (this method and text()) are expected to be "simplest case" methods and that any more complex cases should be handled by the user diving into `response.candidates` and handling it themselves, this makes sense.

packages/main/src/requests/response-helpers.ts
```diff
@@ -54,6 +55,30 @@ export function addHelpers(
...
+    if (response.candidates && response.candidates.length > 0) {
+      if (response.candidates.length > 1) {
+        console.warn(
+          `This response had ${response.candidates.length} ` +
+            `candidates. Returning text from the first candidate only. ` +
```
hsubox76, line 63: text => function call

packages/main/src/requests/request-helpers.ts
```diff
@@ -38,7 +39,52 @@ export function formatNewContent(
...
+
+/**
+ * When multiple Part types (i.e. FunctionResponsePart and TextPart) are
+ * passed in a single Part array, we may need to assign different roles to each
+ * part. Currently only FunctionResponsePart requires a role other than 'user'.
+ * @private
+ * @param parts Array of parts to pass to the model
+ * @returns Array of content items
+ */
+function assignRoleToPartsAndValidateSendMessageRequest(
```
hsubox76, line 54: I think this (through formatNewContent) may end up being called in the ChatSession constructor when an existing history is passed in which includes `model` turns, so it needs to preserve the `model` role when that happens.

hsubox76 (commented):
Haven't completely read through it yet, just have a few comments so far and will finish the rest shortly.

packages/main/src/requests/request-helpers.ts
alx13: Yeah, you are right. I'm just not sure if `startChat` method should accept anything but `Content` with roles properly assigned.

I would assume the use case for `startChat` with history is only valid if the history is obtained through `getHistory` method (or constructed as a proper Content).

In that case we may drop `InputContent` interface at all. WDYT?

packages/main/src/requests/request-helpers.ts
```diff
@@ -38,7 +39,52 @@ export function formatNewContent(
...
+
+  if (hasUserContent && hasFunctionContent) {
+    throw new GoogleGenerativeAIError(
+      "Within a single message, FunctionResponse cannot be mixed with other type of part in the request for sending chat message.",
+    );
+  }
+
+  if (!hasUserContent && !hasFunctionContent) {
+    throw new GoogleGenerativeAIError(
+      "Within a single message, FunctionResponse cannot be mixed with other type of part in the request for sending chat message.",
```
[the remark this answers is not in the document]
alx13: Fixed

packages/main/src/requests/request-helpers.ts
alx13: Added new validation function for chat history

packages/main/src/requests/response-helpers.ts
alx13: For now I guess let's keep this function as is. When there will be a change with multiple functionCalls we can rethink that thing

packages/main/src/methods/chat-session-helpers.ts
```diff
@@ -0,0 +1,92 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Content, POSSIBLE_ROLES, Part, Role } from "../../types";
+import { GoogleGenerativeAIError } from "../errors";
+
+const VALID_PART_FIELDS: Array<keyof Part> = [
```
hsubox76, line 21: Maybe add a comment with a link to https://source.corp.google.com/piper///depot/google3/google/ai/generativelanguage/v1main/content.proto as a source of truth, unfortunately I don't think there's a public version?

packages/main/src/methods/chat-session-helpers.ts
```diff
@@ -0,0 +1,92 @@
...
+  "functionResponse",
+];
+
+const VALID_PARTS_PER_ROLE: { [key in Role]: Array<keyof Part> } = {
+  user: ["text", "inlineData"],
+  function: ["functionResponse"],
+  model: ["text", "functionCall"],
+};
+
+const VALID_PREVIOUS_CONTENT_ROLES: { [key in Role]: Role[] } = {
```
hsubox76, line 34: Is this documented anywhere or do we just know this by trial and error?

packages/main/src/methods/chat-session-helpers.ts
```diff
@@ -0,0 +1,92 @@
...
+  let prevContent: Content;
+  for (const currContent of history) {
+    const { role, parts } = currContent;
+    if (!prevContent && role !== "user") {
+      throw new GoogleGenerativeAIError(
+        "First content should be with role 'user'",
+      );
+    }
+    if (!POSSIBLE_ROLES.includes(role)) {
+      throw new GoogleGenerativeAIError("Each item should include role field");
```
hsubox76, line 50: Can we list the valid roles here? Like
```
`Each item should include a role field. Valid roles: ${POSSIBLE_ROLES.join(',')}`
```
or something. In case they did include a role but they mistyped it or it doesn't exist (like "system" at the moment).

packages/main/src/requests/request-helpers.ts
```diff
@@ -38,7 +38,52 @@ export function formatNewContent(
       }
     }
   }
-  return { role, parts: newParts };
+  // return { role, parts: newParts };
```
hsubox76, line 41: Can be removed now.

packages/main/src/methods/chat-session-helpers.test.ts
```diff
@@ -0,0 +1,153 @@
+/**
+ * @license
+ * Copyright 2023 Google LLC
```
hsubox76, line 3: 2024 here and other new files.

packages/main/types/requests.ts
```diff
@@ -114,3 +117,136 @@ export interface RequestOptions {
...
+
+/**
+ * Structured representation of a function declaration as defined by the
+ * [OpenAPI 3.0 specification](https://spec.openapis.org/oas/v3.0.3). Included
+ * in this declaration are the function name and parameters. This
+ * FunctionDeclaration is a representation of a block of code that can be used
+ * as a Tool by the model and executed by the client.
+ * @public
+ */
+export declare interface FunctionDeclaration {
```
hsubox76, line 134: Can we lift this to the level of `Tool` as described in https://docs.google.com/document/d/1Z0xNfOoUs8sIwSpJjAPNBvL5jVmM-kDJ1MmElfCk2Wc/edit?tab=t.0#heading=h.x9snb54sjlu9 ?

hsubox76: Note: we decided not to do this, we'll leave it as is.

packages/main/types/content.ts
```diff
@@ -54,6 +54,55 @@ export interface TextPart {
 export interface InlineDataPart {
   text?: never;
   inlineData: GenerativeContentBlob;
+  functionCall?: never;
+  functionResponse?: never;
+}
+
+/**
+ * Content part interface if the part represents FunctionResponse.
```
hsubox76, line 62: Reminder to run `yarn docs` to generate reference md files with these new interfaces.

hsubox76:
> Will it be possible to use this feature of function calling along with the multi step (i.e. ChatSession) capabilities? e.g. to call the endpoint .startChat along with an API schema for the model to do function calling?

This PR is for basic function calling, automated function calling will come very soon after.

packages/main/src/methods/chat-session-helpers.ts
alx13: The are public docs on the rest API:
https://ai.google.dev/api/rest/v1beta/Content

lahirumaramba (approved):

packages/main/src/methods/chat-session-helpers.ts
alx13: Partially here: https://ai.google.dev/docs/function_calling#expandable-5

Idea that user/model and function parts should always be interleaved.
I.e. you can get two parts of the same type after each other. 

And you should call a `function` and provide `functionResponse` straight after model gave you answer. (Although you can ignore `functionCall` and make a `user` content.

API will raise an error if request is not formatted properly.

hsubox76 (approved):

PR #50, opened 2024-02-02
feat: add request signal configuration

### LINKED ISSUE #48: Feature Request: signal option
can we add a new `signal` option? it will be more flexible and can be used to abort the stream when a user cancels chat sessions. it meets the timeout requirement as well, `timeout` is also very useful as it's more convenient BTW :)

```ts
const model = genAI.getGenerativeModel({ model: "gemini-pro" }, {
  signal: AbortSignal.timeout(10000),
});
```

https://github.com/google/generative-ai-js/blob/d8f22d2259c9e926dcc4ac955db896c235e0a3bc/packages/main/src/requests/request.ts#L103-L117

_Originally posted by @tmkx in https://github.com/google/generative-ai-js/pull/31#issuecomment-1920556402_

WillianAgostini:
Fixes #48

> Adding new `signal` options

- [x] Tests pass
- [x] Appropriate changes to documentation are included in the PR


packages/main/src/requests/request.ts
```diff
@@ -107,6 +107,9 @@ export async function makeRequest(
  */
 function buildFetchOptions(requestOptions?: RequestOptions): RequestInit {
   const fetchOptions = {} as RequestInit;
+  if (requestOptions?.signal) {
+    fetchOptions.signal = requestOptions?.signal;
```
tmkx, line 111: the optional chaining can be omitted

tmkx (commented):
thank you!

packages/main/src/requests/request.ts
WillianAgostini: Ok, https://github.com/google/generative-ai-js/pull/50/commits/904795ee86436348cccf34d9ca8c761500135771

gaodeng (approved):

gaodeng:
Could you please take a look at this PR, @hsubox76? Thanks in advance!

hsubox76:
Sorry I haven't gotten back to this. One question: what's the expected behavior if both `timeout` and `signal` are specified? It seems like it might be unexpected behavior as is. We should probably log a warning?

Also heads-up, there is a plan to change `timeout` to `timeoutMillis` or `timeoutMilliseconds` so we might end up doing that all in one PR.

tmkx:
There is a real-world library([axios](https://github.com/axios/axios)) that supports both options simultaneously:

```ts
import axios from 'axios';

axios
  .get('https://httpbin.org/delay/2', {
    // catch AxiosError: timeout of 1000ms exceeded
    timeout: 1000,
    // catch CanceledError: canceled
    signal: AbortSignal.timeout(1000),
  })
  .then((response) => console.log('then', response.data))
  .catch((reason) => console.error('catch', reason));
```

hsubox76:
I thought that was actually going to be pretty complex but it looks like we can just collect all the signals the user wants to use (up to 2 in this case) and wrap them with `AbortSignal.any()`: https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal#aborting_a_fetch_with_timeout_or_explicit_abort

hsubox76:
An additional wrinkle is that we don't want an abort signal to be set at the model level, you want to set it on a per-request level, otherwise it cancels all requests. We're working on another PR to do this.

PR #56, opened 2024-03-04
Allow users to choose API endpoint version

hsubox76:
Add an option to `RequestOptions` so that users can choose API version. Right now the options are "v1" (default) and "v1beta".

- Added option
- Updated docs
- Fixed some integration tests

See design doc (Google internal link only) https://docs.google.com/document/d/17EccPH51QeayXPWdDx6Snz5NOLJXm4La41eEFyTXDz8/edit?tab=t.0#heading=h.ymlc40jj6ff0


packages/main/test-integration/web/index.test.ts
```diff
@@ -150,11 +168,7 @@ describe("startChat", function () {
     const question1 = "What is the capital of Oregon?";
     const question2 = "How many people live there?";
     const question3 = "What is the closest river?";
-    const chat = model.startChat({
-      generationConfig: {
-        maxOutputTokens: 100,
-      },
```
lahirumaramba, line 156: curious, is there a reason to remove the max tokens limit here?

packages/main/src/requests/request.ts
```diff
@@ -43,9 +43,11 @@ export class RequestUrl {
     public task: Task,
     public apiKey: string,
     public stream: boolean,
+    public requestOptions: RequestOptions,
   ) {}
   toString(): string {
-    let url = `${BASE_URL}/${API_VERSION}/${this.model}:${this.task}`;
+    const apiVersion = this.requestOptions?.apiVersion || DEFAULT_API_VERSION;
```
lahirumaramba, line 49: Should we validate `this.requestOptions?.apiVersion` to check if it is either `v1` or `v1beta`?

lahirumaramba (approved):
Added two comments. LGTM thanks!

packages/main/test-integration/web/index.test.ts
hsubox76: Yeah, so this is a live backend test so it doesn't always come back with the same results so it's very loose in what it considers a success (any non-empty string). Lately, it will come back with nothing (empty string) if I set the token limit to 100, and I'm not sure any number guarantees something will come back, so I just removed it.

packages/main/src/requests/request.ts
hsubox76: No, I want to future-proof it so any version is allowed, so we don't have to update the SDK every time a new version becomes available on the backend.

packages/main/src/methods/count-tokens.ts
```diff
@@ -28,7 +28,7 @@ export async function countTokens(
   params: CountTokensRequest,
   requestOptions?: RequestOptions,
 ): Promise<CountTokensResponse> {
-  const url = new RequestUrl(model, Task.COUNT_TOKENS, apiKey, false);
+  const url = new RequestUrl(model, Task.COUNT_TOKENS, apiKey, false, {});
```
cketcham, line 31: Shouldn't this be the requestOptions? Otherwise the version won't get put in the url correctly.

PR #62, opened 2024-03-14
validateChatHistory is now checking that 'parts' property is an array

alx13:


.changeset/lazy-houses-attend.md
```diff
@@ -0,0 +1,5 @@
+---
+"@google/generative-ai": patch
+---
+
+validateChatHistory is now checking that 'parts' propery is an array
```
hsubox76, line 5: nit: "property" spelling

packages/main/src/methods/chat-session-helpers.test.ts
```diff
@@ -102,6 +102,16 @@ describe("chat-session-helpers", () => {
         ],
         isValid: true,
       },
+      {
+        //@ts-expect-error
```
hsubox76, line 106: Did not know about this TS annotation before, this is helpful.

hsubox76 (approved):

PR #73, opened 2024-03-29
Add option to modify baseUrl

hsubox76:
Add `baseUrl` property to `RequestOptions` allowing user to set a different base url.

Fixes https://github.com/google/generative-ai-js/issues/25


lahirumaramba (approved):
LGTM! though I am a bit unclear on the use-cases here...

hsubox76:
We discussed this and it seems like there's a few use cases, the most recent being that internal teams want to be able to switch to testing against staging endpoints, etc. Externally there are some reasonable proxy use cases where the developer might have some middleware on their own servers between the client and the google servers, for rate-limiting, tracking, inserting additional prompt content, etc

PR #75, opened 2024-03-29
Deprecate functionCall() and add functionCalls()

hsubox76:
Gemini SDK teams request we do this to maintain alignment across SDKs. Deprecating instead of removing to minimize breaking. Will remove with next major release.


lahirumaramba (approved):
LGTM! Thanks!

PR #90, opened 2024-04-10
DOM text reinterpreted as HTML  shared.js 

Shivam7-1:
By using innerText, it will avoid the risk of HTML injection, as these properties automatically escape any HTML special characters in the provided text. This helps prevent cross-site scripting (XSS) vulnerabilities by treating the input as plain text rather than interpreted HTML. Always be cautious when dealing with user input or dynamic content to prevent security risks.

### DIFF
--- a/samples/web/utils/shared.js
+++ b/samples/web/utils/shared.js
@@ -79,7 +79,7 @@ export async function updateUI(resultEl, getResult, streaming) {
         // Get first candidate's current text chunk
         const chunkText = chunk.text();
         text += chunkText;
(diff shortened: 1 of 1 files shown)


Shivam7-1:
Hi @hsubox76 Could You Please Review this PR
Thanks

hsubox76:
This file isn't part of our source code, it's in the samples directory as an example snippet of how to use the SDK. The parts that are unrelated to usage of the SDK are kept as simple as possible since that's not what the sample is meant to focus on.

PR #91, opened 2024-04-10
{patch} DOM text reinterpreted as HTML Added encodeURIComponent

Shivam7-1:
By using encodeURIComponent(), it will prevent the risk of unintended interpretation of special characters in the URL construction. This method automatically encodes special characters, ensuring that they are treated as part of the URL string rather than being interpreted as HTML or JavaScript code. This helps prevent potential security vulnerabilities, such as injection attacks or unexpected behavior, by encoding the input as part of the URL path.

### DIFF
--- a/samples/web/index.html
+++ b/samples/web/index.html
@@ -68,7 +68,7 @@
         thumbnailsEl.innerHTML = "";
         for (const file of fileInputEl.files) {
           const url = URL.createObjectURL(file);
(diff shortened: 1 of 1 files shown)


Shivam7-1:
Hi @hsubox76 Could You Please Review this PR
Thanks

hsubox76:
This file isn't part of our source code, it's in the samples directory as an example snippet of how to use the SDK. The parts that are unrelated to usage of the SDK are kept as simple as possible since that's not what the sample is meant to focus on.

PR #94, opened 2024-04-11
Change default apiVersion to "v1beta"

hsubox76:
Setting the default apiVersion to "v1beta" matches the other GoogleAI server SDKs (Go and Python).

For web users, this will be slightly different from other client SDKs such as Swift and Kotlin, which default to "v1".

### DIFF
--- a/.changeset/few-crabs-crash.md
+++ b/.changeset/few-crabs-crash.md
@@ -2,4 +2,4 @@
 "@google/generative-ai": minor
 ---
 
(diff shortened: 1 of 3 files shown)


LindaLawton:
can you tell me then how to access the file upload and file list methods? using this [Prompting with media files](https://ai.google.dev/tutorials/prompting_with_media) and here [File prompting strategies](https://ai.google.dev/docs/file_prompt_strategies)

PR #103, opened 2024-04-16
Fix a couple bugs where not all params are passed through to the request

### LINKED ISSUE #98: Request options are not passed to modelInstance.countTokens
My understanding is that the library requires `apiVersion` to be set to `v1beta` for Gemini 1.5 to work. However, the `apiVersion` is not passed down to the implementation of `countTokens`, due to which it makes a request at the older endpoint.

This is with v0.6.0 (latest)

## Actual Behavior

```js
const modelInstance = googleAI.getGenerativeModel(
    { model: "gemini-1.5-pro-latest" },
    { apiVersion: "v1beta" },
  )

modelInstance.generateContent()
// this sends a request to https://generativelanguage.googleapis.com/v1beta/models/gemini-1.5-pro-latest:generateContent
// ...

modelInstance.countTokens
// this sends a request to https://generativelanguage.googleapis.com/v1/models/gemini-1.5-pro-latest:countTokens()
// which errors out
```

I believe the issue lies here (`requestOptions` are not passed): https://github.com/google/generative-ai-js/blob/79b7651547536183108f0b78791761d1ece01f88/packages/main/src/models/generative-model.ts#L156

## Expected Behavior

The `countTokens` method should make a request to the v1beta endpoint

### LINKED ISSUE #101: `ChatSession` don't inherit `generationConfig` from `GenerativeModel`
## Actual Behavior

The [docs](https://ai.google.dev/tutorials/get_started_node#model-parameters) mention that you can pass model parameters to `genAI.getGenerativeModel.getGenerativeModel`:

```
const generationConfig = {
  stopSequences: ["red"],
  maxOutputTokens: 200,
  temperature: 0.9,
  topP: 0.1,
  topK: 16,
};

const model = genAI.getGenerativeModel({ model: "MODEL_NAME",  generationConfig });
```

But if you pass the parameters there, they are silently ignored when using [`model.startChat` API](https://ai.google.dev/tutorials/get_started_node#multi-turn-conversations-chat). You need to pass these parameters directly to `startChat`, even though it is already a method of `GenerativeModel`.

I confirmed that by making the requests through a proxy (https://httptoolkit.com/).

The same applies for `safetySettings`.

## Expected Behavior

`model.startChat` should inherit `generationConfig`, `safetySettings`, etc., from `model`.

## Specifications

- Version: 0.6.0
- Platform: Windows

hsubox76:
Fixed bugs where
- `RequestOptions` was not passed to `countTokens` and `embedContent`
- `generationConfig` and `safetySettings` were not passed through `startChat` to `ChatSession`.

Fixes #101 
Fixes #98

### DIFF
--- a/.changeset/many-chefs-appear.md
+++ b/.changeset/many-chefs-appear.md
@@ -1,0 +1,5 @@
+---
+"@google/generative-ai": patch
+---
+
+Fixed bugs where `RequestOptions`, `generationConfig`, and `safetySettings` were not passed from the model down to some methods.
--- a/packages/main/src/models/generative-model.test.ts
+++ b/packages/main/src/models/generative-model.test.ts
@@ -46,21 +46,27 @@ describe("GenerativeModel", () => {
     expect(genModel.model).to.equal("tunedModels/my-model");
   });
   it("passes params through to generateContent", async () => {
-    const genModel = new GenerativeModel("apiKey", {
-      model: "my-model",
-      generationConfig: { temperature: 0 },
-      safetySettings: [
-        {
-          category: HarmCategory.HARM_CATEGORY_DANGEROUS_CONTENT,
-          threshold: HarmBlockThreshold.BLOCK_LOW_AND_ABOVE,
+    const genModel = new GenerativeModel(
+      "apiKey",
+      {
+        model: "my-model",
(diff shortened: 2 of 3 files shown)


lahirumaramba (approved):
Thanks!

PR #104, opened 2024-04-16
Add Github release workflow

hsubox76:
Add a workflow to run a github release when a release tag is pushed to the repo by release.yml.

I could have added this as a step in release.yml but it's a lot easier and less error-prone to get the tag name this way (when triggered by a tag push, the tag is the github.ref_name)


.github/workflows/gh-release.yml
```diff
@@ -0,0 +1,25 @@
...
+      - name: Create Github release
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          tag: ${{ github.ref_name }}
+        run: |
+          gh release create "$tag" \
+              --repo="$GITHUB_REPOSITORY" \
+              --title="$tag" \
+              --generate-notes
+              --verify-tag
```
lahirumaramba, line 25: nit: new line

lahirumaramba (approved):
Thanks!

.github/workflows/gh-release.yml
hsubox76: Added!

PR #110, opened 2024-04-22
Fix Github release script

hsubox76:
Trigger on all tag pushes. We should only be pushing tags on releases but I added a conditional on the job to make sure the tag matches the pattern.


lahirumaramba (approved):
LGTM! Thanks!

Issue #82, opened 2024-04-03
Issue: Improve Error Handling and Provide Detailed Error Messages


thenameisajay:
Firstly, Thanks for the work on this library :)

### Expected Behavior

When an error occurs while using the API, the error message should be structured in a way that makes it easy to extract the reason for the failure. Currently, the error message is all squashed into one, making it tedious to get the `reason` property from the object, as shown in the following image:

![Error Message](https://github.com/google/generative-ai-js/assets/113372062/fe5b007f-63b4-4aba-9561-ddc6cdb6bc73)

Additionally, since the error is of the type `any`, it makes it difficult to tap into the reason for the failure, as shown in the following image:

![Error Type is any](https://github.com/google/generative-ai-js/assets/113372062/bced0168-a01b-48a0-b26e-380f3dd60e95)

### Workaround

To work around this issue, the user has to manually parse the error message to extract the reason for the failure, as shown in the following code snippet:

```typescript
const makeTheCall = async (
    message: object,
    apikey: string,
    generationConfig: object,
) => {
    await run(message, apikey as string, generationConfig)
        .then((response) => {
            settledState(response as string);
        })
        .catch((error) => {
            if (error.message.includes('[GoogleGenerativeAI Error]')) {
                const errorMessage = error.message;

                if (errorMessage.includes('[400 Bad Request]')) {
                    // Extract the reason for the error
                    const reasonStart = errorMessage.indexOf('reason":"') + 9;
                    const reasonEnd = errorMessage.indexOf('","domain');
                    const reason = errorMessage.substring(
                        reasonStart,
                        reasonEnd,
                    );

                    console.log('Error reason:', reason);
                }
            }
            // errorState(error.reason as string);
            throw new Error(error.message as string);
        });
};
```

This workaround allows the user to extract the reason for the failure, as shown in the following image:

![Error Reason Console logs](https://github.com/google/generative-ai-js/assets/113372062/2eb1f0a8-9e6d-402f-82bc-9117dc566107)

### Proposed Solution

To improve the error handling and provide more detailed error messages, I suggest the following:

1. Implement a custom error class that extends the built-in `Error` class and includes properties for the error code, reason, and any other relevant information.
2. When an error occurs, create an instance of the custom error class and throw it instead of the raw error object.
3. Update the error handling logic in the client code to handle the custom error class, making it easier to extract the relevant information from the error.

This approach will provide a more structured and informative error handling experience for the users of the library.

### Repo Link

[...]

PR #112, opened 2024-04-26
Add more properties to errors and export errors.

hsubox76:
Fixes https://github.com/google-gemini/generative-ai-js/issues/82

- Creates a new `GoogleGenerativeAIFetchError` that extends `GoogleGenerativeAIError` with additional properties:
  - `status` - http status code
  - `statusText` - http status text
  - `errorDetails` - if the error response contained a json with an `error.details` field, which may include fields such as type and reason

- Export all error classes.


packages/main/src/errors.ts
```diff
@@ -31,3 +40,30 @@ export class GoogleGenerativeAIResponseError<
     super(message);
   }
 }
+
+/**
+ * Error class covering http errors when calling the server. Includes http
```
DellaBitta, line 45: We seem to capitalize `http` in our Firebase JS SDK's docs:

[See here](https://firebase.google.com/docs/reference/node/firebase.functions.HttpsError#code).

dlarocque (approved):

DellaBitta (approved):

PR #128, opened 2024-05-03
Add generated docs for custom headers request options

dlarocque:
Add generated docs for custom headers request options.


hsubox76 (approved):
Thanks, LGTM!

PR #129, opened 2024-05-03
Add docs check workflow

hsubox76:
Checks if reference docs need to be generated but haven't been.

Copied and modified from https://github.com/firebase/firebase-js-sdk/blob/master/.github/workflows/check-docs.yml


dlarocque (approved):
I needed this :^)

PR #134, opened 2024-05-09
Support File URLs

### LINKED ISSUE #130: File URL option for the GenAI File API
At the moment a file path has to be utilised: ``mimeType: mime.lookup(filePath),``. However it would be helpful if a file URL could also be used instead of a local file path: ``mimeType: mime.lookup(rawfileURL),``

lahirumaramba:


lahirumaramba:
The FR was on supporting http urls. Closing this PR for now.

PR #135, opened 2024-05-09
Fix bug when no file name is provided for upload

hsubox76:
A bug in the logic caused the FileMetadata.name field to be filled with `files/undefined` instead of actually being left undefined when the user does not provide that value.


lahirumaramba (approved):
LGTM!

PR #139, opened 2024-05-15
Remove chat session's previous content role validation checks

DellaBitta:
The service used to require chat session roles to conform to a certain pattern. This restriction has been lifted and so we shouldn't restrict it on the client-side.

Additionally update the tests around this feature.


hsubox76 (approved):
Looks good! Thanks for adding another test to make sure it errors if "user" isn't first.

PR #144, opened 2024-05-20
AbortSignal support

DellaBitta:
Adds `SingleRequestOptions` with `AbortSignal` support to most of the asynchronous methods of:

 - `GenerativeModel`
 - `GoogleAIFileManager`
 - `ChatSession`

**NOTE:** `AbortSignal` is a client-only operation. Using it to cancel an operation will not cancel the request in the service. You will still be charged usage for any applicable operations.

Some methods are not supported as they would leave the server in an ambiguous state, namely `GoogleAIFileManager.uploadFile()` and `GoogleAIFileManager.deleteFile()`.


packages/main/src/files/file-manager.ts
```diff
@@ -42,23 +42,36 @@ export interface UploadMetadata {
  * @public
  */
 export class GoogleAIFileManager {
+  private _requestOptions: RequestOptions;
   constructor(
     public apiKey: string,
-    private _requestOptions?: RequestOptions,
-  ) {}
+    requestOptions?: RequestOptions,
```
hsubox76, line 48: I don't think we should have a type that says they can pass an abortSignal property but then we immediately delete it, we should probably make a new interface, maybe GlobalRequestOptions or something. Or rename the requestOptions that can be passed to individual methods as SingleRequestOptions or something.

For non-TS users, I think it may helpful to have a warning here if they accidentally provide it. I don't think we should delete it silently. You'll have to wrestle TS a little to provide the warning if you separate into two types but I think that's reasonable as it's to accommodate non-TS users.

packages/main/src/files/file-manager.ts
```diff
@@ -42,23 +42,36 @@ export interface UploadMetadata {
...
   /**
    * Upload a file
    */
   async uploadFile(
     filePath: string,
     fileMetadata: FileMetadata,
+    requestOptions?: RequestOptions,
   ): Promise<UploadFileResponse> {
     const file = readFileSync(filePath);
+    const filesRequestOptions = Object.create(this._requestOptions);
```
hsubox76, line 66: I think you can `const filesRequestOptions = {...this._requestOptions, ...requestOptions}`? I think you can set the requestOptions param to default to `{}` (`requestOptions: RequestOptions = {}`) so the second spread will never fail.

packages/main/src/files/request.ts
```diff
@@ -131,13 +131,23 @@ export async function makeFilesRequest(
...
-    return signal;
+  if (
+    requestOptions?.abortSignal !== undefined ||
+    requestOptions?.timeout >= 0
+  ) {
+    const controller = new AbortController();
+    if (requestOptions?.timeout >= 0) {
+      setTimeout(() => controller.abort(), requestOptions.timeout);
+    }
+    if (requestOptions.abortSignal) {
```
hsubox76, line 146: For reference, here's the previous PR and the discussion on it: https://github.com/google-gemini/generative-ai-js/pull/50

I think we can just keep the previous "signal" in our code (and name it `timeoutSignal`) and then we can just `return AbortSignal.any([timeoutSignal, userSignal])`. We may also want to name the user-provided signal field just "signal" so that it matches the native API, indicating to the user that we're just passing it through with no modifications.

packages/main/test-integration/resources/cat.jpg
hsubox76, line 1: FYI there's a cat.jpeg and other resource files in `test-utils`

packages/main/types/requests.ts
```diff
@@ -147,6 +147,11 @@ export interface RequestOptions {
    * Custom HTTP request headers.
    */
   customHeaders?: Headers | Record<string, string>;
+  /**
+   * An object that may be used to abort aynchronous requests. This is only available on
```
hsubox76, line 151: If we separate RequestOptions into two types, we don't need the "and is ignored by models..." part, although the "only available" part might be helpful. We should also add how this works together with timeout (whichever one triggers first). Also spelling on "asynchronous"

packages/main/src/files/file-manager.ts
DellaBitta: Done.

packages/main/test-integration/resources/cat.jpg
DellaBitta: Done

packages/main/src/files/file-manager.ts
DellaBitta: I couldn't find a proper name to replace `RequestOptions` which encompasses `FileManager`, `ChatSession`,  and `GenerativeModel` which all use `RequestOptions` so I've made a `SingleRequestOptions` instead.

packages/main/types/requests.ts
DellaBitta: Done.

packages/main/src/files/request.ts
DellaBitta: `AbortSignal.any()` requires node v20.3.0, and has only made it into Chrome last fall, and Safari this March. 

Currently our package.json engines directive is `">=18.0.0"`.

DellaBitta: > We may also want to name the user-provided signal field just "signal" so that it matches the native API, indicating to the user that we're just passing it through with no modifications.

Done!

packages/main/test-integration/node/abort-signal.test.ts
```diff
@@ -17,59 +17,59 @@
...
+import { SingleRequestOptions } from "../..";
 import { GoogleAIFileManager } from "../../dist/files";
 
 use(chaiAsPromised);
 
 /**
  * Integration tests against live backend.
  */
-describe("abortSignal", function () {
+describe("signal", function () {
```
hsubox76, line 28: Can you add a test to see if this works together with timeout (e.g., whichever comes first works)? Or if it's too hard to make an automated test, can you test it manually?

docs/reference/files/generative-ai.googleaifilemanager.deletefile.md
```diff
@@ -4,20 +4,22 @@
 
 ## GoogleAIFileManager.deleteFile() method
 
-Delete file with given ID
+Delete file with given ID.
+
+Any fields set in the optional [SingleRequestOptions](./generative-ai.singlerequestoptions.md) parameter will take predendence over the [RequestOptions](./generative-ai.requestoptions.md) values provided at the time of the [GoogleAIFileManager](./generative-ai.googleaifilemanager.md) initialization.
```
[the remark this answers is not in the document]
DellaBitta: Copy pasta! Thanks, and fixed.

packages/main/test-integration/node/abort-signal.test.ts
DellaBitta: Done.

gaodeng (approved):

gaodeng:
any news?

hsubox76 (approved):
Can you add a changeset? Minor because we're adding to the API.

PR #148, opened 2024-05-22
Add GenerateContentRequest as an optional param to CountTokensRequest

DellaBitta:
Expand the model's countTokens method to alternatively accept a GenerateContentRequest.

Added integration and unit tests.


packages/main/types/requests.ts
```diff
@@ -99,7 +107,17 @@ export interface StartChatParams extends BaseParams {
  * @public
  */
 export interface CountTokensRequest {
-  contents: Content[];
+  generateContentRequest?: GenerateContentRequest;
+  contents?: Content[];
+}
+
+/**
+ * Params for calling {@link GenerativeModel.countTokens}
```
hsubox76, line 115: This comment should probably be updated.

packages/main/src/requests/request-helpers.ts
```diff
@@ -104,6 +106,31 @@ function assignRoleToPartsAndValidateSendMessageRequest(
   return functionContent;
 }
 
+export function formatCountTokensInput(
+  params: CountTokensRequest | string | Array<string | Part>,
+  model: string,
+): CountTokensRequestInternal {
+  let formattedRequest: CountTokensRequestInternal = {};
+  const containsGenerateContentRequest =
+    (params as CountTokensRequest).generateContentRequest !== undefined;
```
hsubox76, line 115: Could use `!= null` which includes undefined, in case the user has decided to do weird things and null it out or something.

packages/main/src/requests/request-helpers.ts
```diff
@@ -104,6 +106,31 @@ function assignRoleToPartsAndValidateSendMessageRequest(
...
+  if ((params as CountTokensRequest).contents) {
+    if (containsGenerateContentRequest) {
+      throw new GoogleGenerativeAIError(
```
hsubox76, line 118: This is a great place to use the brand new `GoogleGenerativeAIRequestInputError` we are using for bad input. https://github.com/google-gemini/generative-ai-js/blob/main/packages/main/src/errors.ts#L66 Although I am not 100% sure we throw in this situation. Do you want to double check in Gemini chat? What does the backend do? Does it error? Does it ignore one? Which one?

packages/main/types/requests.ts
```diff
@@ -99,7 +107,17 @@ export interface StartChatParams extends BaseParams {
  * @public
```
hsubox76, line 107: If you resolve the question of what happens and what's supposed to happen if you supply both, we should document it here (it throws an error, it prefers one or the other, it combines them in some weird way, or whatever the answer is)

packages/main/types/requests.ts
DellaBitta: Done.

packages/main/types/requests.ts
DellaBitta: Consulted the group and an error when both are defined is preferred.

packages/main/src/requests/request-helpers.ts
DellaBitta: Done. And team consensus is to throw an error here (I checked).

packages/main/src/requests/request-helpers.ts
DellaBitta: Done.

.changeset/dirty-wolves-sin.md
```diff
@@ -0,0 +1,5 @@
+---
+"@google/generative-ai": minor
+---
+
+Expand the model's countTokens method to alternatively accept a GenerateContentRequest.
```
hsubox76, line 5: Nit: backticks on `countTokens` and `GenerateContentRequest`

packages/main/src/models/generative-model.ts
```diff
@@ -157,7 +158,8 @@ export class GenerativeModel {
   async countTokens(
     request: CountTokensRequest | string | Array<string | Part>,
   ): Promise<CountTokensResponse> {
-    const formattedParams = formatGenerateContentInput(request);
+    const formattedParams = formatCountTokensInput(request, this.model);
+    // countTokens() appends the model to the CountTokensRequest proto.
```
hsubox76, line 162: I think this comment should be for the previous line? `countTokens` doesn't append anything. It uses the model param to build the request url.

DellaBitta: It adds it to the request body, but I'll remove the comment.

.changeset/dirty-wolves-sin.md
DellaBitta: Done.

packages/main/src/models/generative-model.ts
hsubox76: countTokens() doesn't, but formatCountTokensInput() does, right? That's why I think it would be best to attach it to the previous line? countTokens() just calls makeRequest() and returns response.json().

DellaBitta: countTokens() also adds it to the proto request object [here](https://github.com/google-gemini/generative-ai-js/blob/9318ee8d5aa0d15a65039af37a9d972af0739b10/packages/main/src/methods/count-tokens.ts#L36).

hsubox76: Hm, that seems to have been there since initial commit but I can't remember why it was needed. I don't know if countTokens is special or if I forgot to remove the model there when I removed it from generateContent and the other methods, let me double check. If it's just left over from the beginning we should probably remove it to avoid unexpected behavior.

hsubox76: I don't think this blocks this PR though.

hsubox76 (approved):

Issue #155, opened 2024-05-27
Cannot find module '@google/generative-ai/files' or its corresponding type declarations.


duccdev:
### Description of the bug:

Hi everyone, I'm trying to import the `GoogleAIFileManager` class from `@google/generative-ai/files`
However, VSCode and the TypeScript compiler itself throws errors about the module not existing
I can import everything from `@google/generative-ai` but not `@google/generative-ai/files`

### Actual vs expected behavior:

This should import the GoogleAIFileManager class:
```ts
import { GoogleAIFileManager } from "@google/generative-ai/files";
```
I get this error:
```
Cannot find module '@google/generative-ai/files' or its corresponding type declarations.
```

### Any other information you'd like to share?

I peeked into the `node_modules` (specifically `node_modules/@google/generative-ai/files/package.json`)
On this specific line:
```json
  "name": "@google/generative-ai/files",
```
I get this warning from VSCode:
```
String does not match the pattern of "^(?:(?:@(?:[a-z0-9-*~][a-z0-9-*._~]*)?/[a-z0-9-._~])|[a-z0-9-~])[a-z0-9-._~]*$".
```
Could this be an issue when importing?

duccdev:
I have found a workaround
Importing stuff from `@google/generative-ai/dist/src/files` instead of `@google/generative-ai/files` works
You can do this for now while this issue gets fixed

hsubox76:
Thanks for the report, I looked into it and it seems like the name format in `node_modules/@google/generative-ai/files/package.json` is incorrect, but not what's causing the issue. The issue is the paths further down in the file are wrong - should begin with`../` and not `./`. You can actually get around this by changing your tsconfig "module" field to "node16" or similar instead of "commonjs" (which makes it rely on the top-level package.json instead, which has the right paths) but I should be able to fix this quickly.

antoniovini47:
For those who are trying to do that recently and getting the same error. I found that Google made some changes and doesn't update some of the documentations.

You doesn't import anymore like:

`import { GoogleAIFileManager } from "@google/generative-ai/files";`

use

`import { GoogleAIFileManager } from "@google/generative-ai/server";`

instead.

Heres the only place they announced it:

![image](https://github.com/google-gemini/generative-ai-js/assets/7387249/9f7d1aa3-902c-4ea3-9eea-830edbdb8fbd)

antoniovini47:
> when i try npm install @google/generative-ai/server i get module not found?

[...]

PR #156, opened 2024-05-28
Fix @google/generative-ai/files paths

hsubox76:
Fixes https://github.com/google-gemini/generative-ai-js/issues/155

This affects TypeScript users that are using "module: commonjs" or any other resolution that relies on the package.json in the files/ directory instead of on package.json exports.

### DIFF
--- a/packages/main/files/package.json
+++ b/packages/main/files/package.json
@@ -1,8 +1,8 @@
 {
-  "name": "@google/generative-ai/files",
+  "name": "@google/generative-ai-files",
   "description": "GoogleAI file upload manager",
-  "main": "./dist/files/index.js",
-  "browser": "./dist/files/index.mjs",
-  "module": "./dist/files/index.mjs",
-  "typings": "./dist/files/files.d.ts"
+  "main": "../dist/files/index.js",
+  "browser": "../dist/files/index.mjs",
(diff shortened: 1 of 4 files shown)


packages/main/files/package.json
```diff
@@ -1,8 +1,8 @@
 {
-  "name": "@google/generative-ai/files",
+  "name": "@google/generative-ai-files",
   "description": "GoogleAI file upload manager",
-  "main": "./dist/files/index.js",
-  "browser": "./dist/files/index.mjs",
-  "module": "./dist/files/index.mjs",
-  "typings": "./dist/files/files.d.ts"
+  "main": "../dist/files/index.js",
```
lahirumaramba, line 4: Would this break the module imports?

lahirumaramba (changes requested):

packages/main/files/package.json
hsubox76: Module imports? I realized it might affect the rollup config as it gets paths from here, so I changed that.

hsubox76: I tested this with tsconfig set to module:commonjs, module:node16 and in a vite bundler proejct and it seems to work in all cases.

lahirumaramba (approved):
LGTM!

PR #158, opened 2024-05-28
Add `responseSchema` to `generationConfig`

lahirumaramba:
Add `responseSchema` support to `GenerationConfig`
Restructure `FunctionDeclarationSchemaProperty` to extend from a base type `Schema`
Add new `ResponseSchema` type


.changeset/five-spoons-matter.md
```diff
@@ -0,0 +1,5 @@
+---
+"@google/generative-ai": minor
+---
+
+Added `responseSchema` to `GenerationConfig` to allow for JSON schema reponses.
```
hsubox76, line 5: suggest "to allow user to provide a JSON schema when `responseMimeType` is set to JSON."

packages/main/src/models/generative-model.test.ts
```diff
@@ -68,6 +81,15 @@ describe("GenerativeModel", () => {
...
     expect(genModel.generationConfig?.temperature).to.equal(0);
+    expect(genModel.generationConfig?.responseMimeType).to.equal(
+      "application/json",
+    );
+    expect(genModel.generationConfig?.responseSchema.type).to.equal(
+      FunctionDeclarationSchemaType.OBJECT,
+    );
+    expect(
+      genModel.generationConfig?.responseSchema.properties.testField.type,
+    ).to.equal(FunctionDeclarationSchemaType.STRING);
```
hsubox76, line 92: I would like to add a test to ensure this is being passed to makeRequestStub and add this to the test below to ensure override works, as well as to the chat.sendMessage tests. I can make a PR to your branch to do it.

hsubox76 (approved):

packages/main/src/models/generative-model.test.ts
lahirumaramba: Thank you! (addressed in #159)

PR #159, opened 2024-05-29
Add some unit test cases

hsubox76:
For generatecontent, test that the field is passed to makeRequest, that it's overrideable by individual methods, and the same for chat.sendMessage.


lahirumaramba (approved):
Thank you! LGTM!

PR #163, opened 2024-05-31
Clean up iTests for CI

DellaBitta:
Change iTests so that they're ready for CI again.

- Remove tests that detected status failures on harmful content. These questions now have textual responses. I attempted to find another means to detect the response is a rejected one (a code or status in the response) but there doesn't seem to be a metric that flags it as a harmful question.
- Change the wording of one of the questions "How many people live there?" as sometimes it wouldn't pass the service's safety checks. This question is now "How many people live in that city?".

This is a change to our tests only, so no Changeset is required.


hsubox76 (approved):
Thanks! LG!

PR #167, opened 2024-06-06
Add the Content caching feature

hsubox76:
# Feature: Content Caching
Design doc (API) (internal): https://docs.google.com/document/d/1kJTdoz-RO7mkzsjE1QDfJfD1YvSxJXuzmqTAtsS1hJ0/edit?resourcekey=0-cv2C2J8SLDn__yoetTPvBw&tab=t.0
Design doc (Python SDK) (internal): https://docs.google.com/document/d/11en7rPCuXIu-hfF71jUDWkfQctWeFEoq_5KgkinnxWg/edit?resourcekey=0-ntvB0-Y_GBQXOfkO-KDaXg&tab=t.0

## APIs added
- GoogleAICacheManager class (exported from new /server path) with create, update, get, list, delete methods.
- GoogleGenerativeAI.getGenerativeModelFromContextCache() added

## Infrastructure changed
- `/files` subpath changed to `/server`. GoogleAIFileManager and GoogleAICacheManager will be exported from here as they are only meant to be used in a server environment.
- All types moved to a single folder.
- Generate common/api-review md files for easier review of API mistakes/gaps
- Refactor `makeRequest`-related functions to reduce code duplication


packages/main/src/requests/request.ts
```diff
@@ -130,89 +130,86 @@ export async function constructRequest(
...
-        response.status,
-        response.statusText,
-        errorDetails,
-      );
+    err.stack = e.stack;
+  }
+  throw err;
+}
+
+export async function handleResponseNotOk(
```
DellaBitta, line 189: And this?

packages/main/src/server/cache-manager.ts
```diff
@@ -0,0 +1,169 @@
...
+  CachedContentCreateParams,
+  CachedContentUpdateParams,
+  ListCacheResponse,
+  ListParams,
+} from "../../types/server";
+import { RpcTask } from "./constants";
+import { GoogleGenerativeAIError } from "../errors";
+
+/**
+ * Class for managing GoogleAI file uploads.
```
DellaBitta, line 36: ContentCaches, not file uploads.

packages/main/src/server/cache-manager.ts
```diff
@@ -0,0 +1,169 @@
...
+  toolConfig?: ToolConfig;
+  systemInstruction?: string | Part | Content;
+
+  constructor(
+    public apiKey: string,
+    private _requestOptions: RequestOptions,
+  ) {}
+
+  /**
+   * Upload a new content cache
```
DellaBitta, line 52: Should be create a new instance of CachedContent, not file upload, and maybe mention that it hits the service..?

packages/main/src/server/cache-manager.ts
```diff
@@ -0,0 +1,169 @@
...
+   */
+  async create(
+    createOptions: CachedContentCreateParams,
+  ): Promise<CachedContent> {
+    const newCachedContent: CachedContent = { ...createOptions };
+    if (createOptions.ttlSeconds) {
+      newCachedContent.ttl = createOptions.ttlSeconds.toString() + "s";
+      delete (newCachedContent as CachedContentCreateParams).ttlSeconds;
+    }
+    if (!newCachedContent.model.includes("/")) {
```
DellaBitta, line 62: What if model is empty, shouldn't we report an error? Or do we leave that up to the service?

packages/main/src/server/request.ts
```diff
@@ -65,17 +54,50 @@ export class FilesRequestUrl {
...
+export class FilesRequestUrl extends ServerRequestUrl {
+  constructor(
+    public task: RpcTask,
+    public apiKey: string,
+    public requestOptions?: RequestOptions,
+  ) {
+    super(task, apiKey, requestOptions);
+    const apiVersion = this.requestOptions?.apiVersion || DEFAULT_API_VERSION;
+    const baseUrl = this.requestOptions?.baseUrl || DEFAULT_BASE_URL;
+    let initialUrl = baseUrl;
```
DellaBitta, line 81: missing an `l`. `initialUrl`.

packages/main/types/function-calling.ts
```diff
@@ -0,0 +1,163 @@
+import { FunctionCallingMode } from "./enums";
```
DellaBitta, line 1: copyright.

packages/main/types/server/caching.ts
```diff
@@ -0,0 +1,72 @@
...
+}
+
+/**
+ * Describes CachedContent interface for sending to the server (if creating)
+ * or received from the server (using getters or list methods).
+ * @public
+ */
+export interface CachedContent extends CachedContentBase {
+  name?: string;
+  ttl?: string;
```
DellaBitta, line 38: Is there a particular format we should define here? Seconds, or something?

packages/main/types/server/caching.ts
```diff
@@ -0,0 +1,72 @@
...
+  // sent field needs to be protobuf Duration ("3.0001s")
+  ttlSeconds?: number;
+}
+
+/**
+ * Params to pass to {@link GoogleAICacheManager.update}
+ * @public
+ */
+export interface CachedContentUpdateParams {
+  cachedContent: CachedContentCreateParams;
```
DellaBitta, line 59: hm, not a fan of this naming convention but I don't konw what to change it to.

packages/main/src/server/request.ts
hsubox76: ? I think I see two Ls?

packages/main/types/server/caching.ts
hsubox76: The input (CachedContentCreateParams) takes `ttlSeconds`. This is the format of the JSON that we send/receive from the API endpoint.

packages/main/src/server/request.ts
DellaBitta: Proof positive that I need glasses.

packages/main/types/function-calling.ts
hsubox76: FYI `yarn format` handles this automatically so we don't need to check for it in reviews.

packages/main/server/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "@google/generative-ai-server",
+  "description": "GoogleAI file upload manager",
```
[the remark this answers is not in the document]
hsubox76: Updated!

packages/main/src/requests/request.ts
```diff
@@ -130,89 +130,86 @@ export async function constructRequest(
...
+  }
+
+  if (!response.ok) {
+    await handleResponseNotOk(response, url);
+  }
+
+  return response;
+}
+
+export function handleResponseError(e: Error, url: string): void {
```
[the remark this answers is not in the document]
hsubox76: Nope, removed export.

packages/main/src/requests/request.ts
hsubox76: Removed export.

packages/main/src/server/cache-manager.ts
hsubox76: Fixed.

packages/main/src/server/cache-manager.ts
hsubox76: I think "upload" is pretty clear that it hits the service? It doesn't create any instance. I notice actually all the variables for headers in this class are called `uploadHeaders` for some reason so I changed them to `headers`. Seems to be a problem with GoogleAIFileManager too but don't want to mess with too many files at once. It's not public facing anyway.

packages/main/src/server/cache-manager.ts
hsubox76: Seems like a good check, added.

packages/main/types/server/caching.ts
DellaBitta: 👍

DellaBitta (approved):

DellaBitta:
Fixes merged.
Documentation merged.

PR #171, opened 2024-06-13
Add Integration Tests for Content Caching

DellaBitta:
Updates to the Content Caching feature in development.

 - adds integration tests
 - adds display name to the creation of content caching.
 - fixes the use of a context cache in the sample implementation.

Documentation PR is separate: #172


packages/main/test-integration/node/cache-content.test.ts
```diff
@@ -0,0 +1,266 @@
...
+import { GoogleGenerativeAI } from "../..";
+import { GoogleAICacheManager } from "../../server"; //= require("@google/generative-ai/server");
+
+use(chaiAsPromised);
+
+/**
+ * Integration tests against live backend.
+ */
+
+describe("cacheContent", function () {
```
dlarocque, line 29: We only have positive test cases. Is it worth testing cases where we expect errors? For example, testing that we get an error if we pass an invalid cached content name with blank spaces, special characters, empty string, etc...
Example in Python SDK: https://github.com/google-gemini/generative-ai-python/pull/355/files#diff-e57945cd50529ced9f0b8aa12022ab5140a6c706d74ecfd05280778f707d3d36R161-R177

packages/main/test-integration/node/cache-content.test.ts
```diff
@@ -0,0 +1,266 @@
...
+  this.slow(10e3);
+  let text: string = "";
+  const endpointConfiguration = { /* remove at launch */ };
+
+  for (let i = 0; i < 1500; i++) {
+    text += "Purple cats drink gatorade.";
+    text += i % 8 === 7 ? "\n" : " ";
+  }
+  it("createCache", async () => {
+    // cacheManager create
```
dlarocque, line 40: We create several cached contents every time we run the integration tests, and we don't delete them. Could this ever cause the tests to break (e.g. usage limits)?

packages/main/test-integration/node/cache-content.test.ts
```diff
@@ -0,0 +1,266 @@
...
+    const result = await model.generateContent({
+      contents: [
+        {
+          role: "user",
+          parts: [{ text: "What do purple cats drink?" }],
+        },
+      ],
+    });
+    const response = await result.response;
+    expect(response.text().toLowerCase().includes("gatorade")).to.be.true;
```
dlarocque, line 264: Could we configure the model to give us more predictable outputs so that this test doesn't flake as often? It looks like Andrew did something like this in the swift SDK https://github.com/google-gemini/generative-ai-swift/blob/main/Tests/GoogleAITests/GoogleAITests.swift#L26

We do tell the model the answer 1500 times so maybe it's not necessary for this test 😅

packages/main/test-integration/node/cache-content.test.ts
DellaBitta: They have a very short TTL.

packages/main/test-integration/node/cache-content.test.ts
DellaBitta: Yeah, I haven't had any flakes in this scernario because the content cache is so specific, but I do argree that we should go back and alter some of our older tests.

packages/main/test-integration/node/cache-content.test.ts
dlarocque: Oh, right!

packages/main/test-integration/node/cache-content.test.ts
DellaBitta: Discussed this offline. There are some unit tests already for negative scenarios where we test the SDK behavior itself. I will work to add some purposely failing integration tests, too, but that shouldn't block this release.

packages/main/src/server/cache-manager.test.ts
```diff
@@ -53,6 +53,7 @@ describe("GoogleAICacheManager", () => {
       systemInstruction: "talk like a cat",
       tools: [{ functionDeclarations: [{ name: "myFn" }] }],
       toolConfig: { functionCallingConfig: {} },
+      displayName: "a display name.",
```
hsubox76, line 56: If adding it to the call, you should probably add an expectation that it's been added to the body, otherwise no reason to add it here (other than to see there's no error). See the other requestBody expectations below.

samples/node/content-caching.js
```diff
@@ -49,9 +49,7 @@ async function run() {
```
hsubox76, line 49: I guess can you add a TODO or reminder to update line 41 here with the new official model name along with where you're going to do it in the integration test?

DellaBitta: I just added it in at this point. Done!

packages/main/src/server/cache-manager.test.ts
DellaBitta: Update these tests.
  - now tests that displayName is in the request.
  - added model as a constant at the top of the test, and use the correct model name for launch. (I know it doesn't matter what the model name is, but it's good to be consistent.)

hsubox76 (approved):

DellaBitta:
Updates to the Content Caching feature in development.

 - adds integration tests
 - adds display name to the creation of content caching.
 - fixes the use of a context cache in the sample implementation.

Documentation PR is separate: #172

PR #178, opened 2024-06-21
UsageMetadata - add cachedContentTokenCount field

DellaBitta:
Add a `cachedContentTokenCount` field to the `UsageMetadata` interface returned by `generateContent` responses.


packages/main/types/responses.ts
```diff
@@ -98,6 +98,8 @@ export interface UsageMetadata {
   candidatesTokenCount: number;
   /** Total token count for the generation request (prompt + candidates). */
   totalTokenCount: number;
+  /** Total token count in the cached part of the prompt, i.e. in the cached content. */
+  cachedContentTokenCount: number;
```
hsubox76, line 102: One thing just occurred to me, are these in fact always populated? (I'm not even sure about the previously existing fields here.) I know the proto doesn't list any of these as optional but it doesn't seem like these would always be populated in actual practice. If not we should probably make it optional. I did a quick test against autopush not using cached content and I got back the 3 other properties but not this one.

DellaBitta: Yeah, it's not always populated. I figured if this was a problem then the non-cached-content integration tests would have failed.

I can make it optional.

DellaBitta: Done!

hsubox76 (approved):

PR #196, opened 2024-07-02
Add canonical textgen samples

hsubox76:
This is part of a larger project to convert samples to a standard format across all languages that can be used in devsite.

See internal doc: https://docs.google.com/spreadsheets/d/10X--DI0YPL3QfEMObe387rYg0t4-kuv33yKn2TIal7E/edit?resourcekey=0-kfy1Ry4YAZh6O4DltLjV9w&gid=0#gid=0

This PR addresses the first file ("text_generation") described in that doc which includes 9 examples. The code to be used in docs is delineated with [START] and [END] region tags marked with the canonical name for that sample.

Media files were taken from the live docs site (https://ai.google.dev/gemini-api/docs/vision?lang=node#technical-details-video) when available and from the Python repo when not available (e.g., audio sample): https://github.com/google-gemini/generative-ai-python/tree/main/third_party

I will delete and replace existing samples files in future PRs as the new samples for those cases are written.


samples/node/text_generation.js
```diff
@@ -0,0 +1,258 @@
...
+    };
+  }
+
+  const prompt = "Describe how this product might be manufactured";
+  // Note: The only accepted mime types are some image types, image/*.
+  const imagePart = fileToGenerativePart("./utils/jetpack.jpg", "image/jpeg");
+
+  const result = await model.generateContentStream([prompt, imagePart]);
+
+  // print text as it comes in
```
DellaBitta, line 96: Add a period. I think this repeats, so I'll stop commenting.

DellaBitta (changes requested):

samples/node/text_generation.js
```diff
@@ -0,0 +1,258 @@
+/**
+ * @license
+ * Copyright 2023 Google LLC
```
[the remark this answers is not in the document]
hsubox76: Done.

samples/node/text_generation.js
hsubox76: Done.

hsubox76:
Made some more changes after doing live tests:

1. Removed `web/` dir - doesn't seem like it's staying up to date, should be easy enough for a web developer to consume Node examples.
2. Use a `mediaPath` path for all media in the samples.
3. Upload video files for inference instead of trying to pass them inline.

DellaBitta (approved):

PR #199, opened 2024-07-03
Add more samples in standardized format

hsubox76:
Adding the following samples in the new standardized sample format.

See https://docs.google.com/spreadsheets/d/10X--DI0YPL3QfEMObe387rYg0t4-kuv33yKn2TIal7E/edit?resourcekey=0-kfy1Ry4YAZh6O4DltLjV9w&gid=0#gid=0 (internal)

filename|region tag
-- | --
function_calling | function_calling
controlled_generation | json_controlled_generation
controlled_generation | json_no_schema
model_configuration | configure_model
safety_settings | safety_settings
safety_settings | safety_settings_multi
system_instruction | system_instruction


samples/node/safety_settings.js
```diff
@@ -0,0 +1,89 @@
...
+
+  const result = await model.generateContent(unsafePrompt);
+
+  try {
+    result.response.text();
+  } catch (e) {
+    console.error(e);
+    console.log(result.response.candidates[0].safetyRatings);
+  }
+  // [END safety_settings]
```
dlarocque, line 49: This throws an error around half the time for me.

samples/node/safety_settings.js
```diff
@@ -0,0 +1,89 @@
...
+    result.response.text();
+  } catch (e) {
+    console.error(e);
+    console.log(result.response.candidates[0].safetyRatings);
+  }
+  // [END safety_settings_multi]
+}
+
+async function run() {
+  // await safetySettings();
```
dlarocque, line 85: I think we want to un-comment this call?

dlarocque (changes requested):

samples/node/safety_settings.js
hsubox76: Oops, fixed.

samples/node/safety_settings.js
hsubox76: I made a modification that seems to trigger it every time I've tried, let me know what you think.

dlarocque (approved):

PR #203, opened 2024-07-08
Add count_tokens sample file

hsubox76:
See spreadsheet (internal) for requirements: https://docs.google.com/spreadsheets/d/10X--DI0YPL3QfEMObe387rYg0t4-kuv33yKn2TIal7E/edit?resourcekey=0-kfy1Ry4YAZh6O4DltLjV9w&gid=0#gid=0

Python version: https://github.com/google-gemini/generative-ai-python/pull/403/files

### DIFF
--- a/samples/node/count_tokens.js
+++ b/samples/node/count_tokens.js
@@ -205,7 +205,7 @@ async function tokensCachedContent() {
   const cacheManager = new GoogleAICacheManager(process.env.API_KEY);
   const cacheResult = await cacheManager.create({
     ttlSeconds: 600,
-    model: "models/gemini-1.5-pro-001",
+    model: "models/gemini-1.5-flash-001",
     contents: [
       {
         role: "user",
@@ -216,7 +216,7 @@ async function tokensCachedContent() {
 
   const genAI = new GoogleGenerativeAI(process.env.API_KEY);
   const model = genAI.getGenerativeModel({
-    model: "models/gemini-1.5-pro-001",
+    model: "models/gemini-1.5-flash",
   });
 
   const result = await model.countTokens({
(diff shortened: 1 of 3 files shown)


samples/node/count_tokens.js
```diff
@@ -0,0 +1,316 @@
...
+  // Count tokens in a prompt without calling text generation.
+  const countResult = await model.countTokens(
+    "The quick brown fox jumps over the lazy dog.",
+  );
+
+  console.log(countResult.totalTokens); // 11
+  console.log(countResult.contentTokens[0]);
+  /**
+   * { partTokens: [ 10 ], roleTokens: 1 }
+   */
```
dlarocque, line 47: Nit: We use a variety of formats for comments that contain JSON. Some use block comments `/* */`, others use multiple single-line comments `//`. Some are formatted multi-line
```
{
  totalTokens: x
  ...
}
```
While others are inline
```
/**
 * { partTokens: x, ... }
 */

**Review by dlarocque — APPROVED** [review]

File: samples/node/count_tokens.js [hunk 2]
```diff
@@ -0,0 +1,316 @@
...
+  let longContentString = "";
+  for (let i = 0; i < 32001; i++) {
+    longContentString += "Purple cats drink lemonade.";
+    longContentString += i % 8 === 7 ? "\n" : " ";
+  }
+
+  const cacheManager = new GoogleAICacheManager(process.env.API_KEY);
+  const cacheResult = await cacheManager.create({
+    ttlSeconds: 600,
+    model: "models/gemini-1.5-pro-001",
```
dlarocque, line 209: Why don't we use 1.5 flash here?

samples/node/count_tokens.js
hsubox76: I meant to convert them all to single line (easier to read, more compact, IDE handles it easily) - fixed this instance.

samples/node/count_tokens.js
hsubox76: I'm not entirely sure why but `gemini-1.5-flash` doesn't work:

GoogleGenerativeAIFetchError: [GoogleGenerativeAI Error]: Error fetching from https://generativelanguage.googleapis.com/v1beta/cachedContents: [400 Bad Request] Model models/gemini-1.5-flash not found or does not support caching. Use ListModels to find a model that supports createCachedContent.

But `gemini-1.5-flash-001` does. I converted it to use that for now and I'm going to ask if this is intended or a bug.

hsubox76: Oh, never mind. See note here. https://ai.google.dev/gemini-api/docs/caching?lang=node So flash-001 should be correct.

PR #205, opened 2024-07-10
Add canonical samples for chat, caching, embeds, and code execution

hsubox76:
Finish required samples (internal link: https://docs.google.com/spreadsheets/d/10X--DI0YPL3QfEMObe387rYg0t4-kuv33yKn2TIal7E/edit?resourcekey=0-kfy1Ry4YAZh6O4DltLjV9w&gid=0#gid=0)

file | region tag
-- | --
chat | chat
chat | chat_streaming
chat | chat_streaming_with_images
cache | cache_create
cache | cache_create_from_name
cache | cache_create_from_chat
cache | cache_delete
cache | cache_get
cache | cache_list
cache | cache_update
embed | embed_content
embed | batch_embed_contents
code_execution | code_execution_basic
code_execution | code_execution_request_override
code_execution | code_execution_chat

Mostly trying to conform to Python: https://github.com/google-gemini/generative-ai-python/tree/main/samples

Also updated text_generation.js to use more convenient ways to display streaming and non streaming responses (reduce lines, write streaming to process.stdout to prevent line breaks).

Next PR is to add a script to add comments to each sample explaining what import lines (e.g., `import { GoogleGenerativeAI } from '@google/generative-ai'`) will need to be added, since we cannot otherwise include import lines in the samples.


samples/node/cache.js
```diff
@@ -0,0 +1,290 @@
...
+              fileUri: uploadResult.file.uri,
+              mimeType: uploadResult.file.mimeType,
+            },
+          },
+        ],
+      },
+    ],
+  });
+  await cacheManager.delete(cacheResult.name);
+  // [END cache_delete]
```
dlarocque, line 174: The other samples in this file `END` the sample before deleting the cached content. Do we want to do this here, too?

samples/node/text_generation.js
```diff
@@ -176,7 +170,7 @@ async function textGenMultimodalMultiImagePromptStreaming() {
   // Print text as it comes in.
   for await (const chunk of result.stream) {
     const chunkText = chunk.text();
-    console.log(chunkText);
+    process.stdout.write(chunkText);
```
dlarocque, line 173: For my own understanding: Why do we need to do `process.stdout.write` instead of `console.log` in streaming samples?

dlarocque (approved):

samples/node/text_generation.js
dlarocque: Ah- sorry, I see your comment about this preventing line breaks.

samples/node/cache.js
hsubox76: No, because this is the delete sample, so we want to show it in the sample.

samples/node/text_generation.js
hsubox76: There's a line break after every console.log but there isn't after process.stdout.write. So this makes the streaming continue the line if the chunk broke in the middle of a sentence or something, instead of starting the next chunk on a newline, so the final output looks normal.

PR #207, opened 2024-07-11
Add import comments to all samples

hsubox76:
ESM requires import statements to only occur once, at the top of each file. Since we have multiple samples in each file, we use comments to indicate to users which imports are needed for each sample, and what the import statements should be.

I wrote a script (`insert-import-comments.js`) to parse each sample and identify which symbols it needs to import, and from which path, and add the comments, and ran the script.

I also abstracted the code that parses function info and shared it between the new script and the check-samples.js script.

### DIFF
--- a/samples/cache.js
+++ b/samples/cache.js
@@ -28,6 +28,9 @@ const mediaPath = __dirname + "/media";
 
 async function cacheCreate() {
   // [START cache_create]
+  // Make sure to include these imports:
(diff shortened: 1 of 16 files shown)


dlarocque (approved):
Very cool script

wwgamorcom (approved):

wwgamorcom (approved):
qqwweerrttyyuuiioo1p1999@gmail.com

PR #208, opened 2024-07-12
Update count_tokens samples

hsubox76:
Update count_tokens samples to better match python: https://github.com/google-gemini/generative-ai-python/blob/main/samples/count_tokens.py

### DIFF
--- a/samples/count_tokens.js
+++ b/samples/count_tokens.js
@@ -255,7 +255,7 @@ async function tokensMultimodalVideoAudioFileApi() {
 async function tokensCachedContent() {
   // [START tokens_cached_content]
   // Make sure to include these imports:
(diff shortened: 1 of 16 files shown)


samples/count_tokens.js
```diff
@@ -143,45 +195,28 @@ async function tokensMultimodalVideoAudioFileApi() {
...
-      }
-      resolve();
-    });
-  }
-
-  const uploadAudioResult = await fileManager.uploadFile(
-    `${mediaPath}/samplesmall.mp3`,
-    { mimeType: "audio/mp3" },
-  );
+  function waitForProcessing(fileName) {}
```
dlarocque, line 198: Can we just remove this empty function?

dlarocque (approved):
Left one comment but approved so we can just merge once it's resolved

PR #209, opened 2024-07-12
Fix count tokens to include model params

hsubox76:
Count tokens request should include any params set on the model parent class, including systemInstruction, tools, cachedContent.

Updated samples to use this.

generationConfig, toolConfig, and safetySettings don't seem like they would affect tokenCount but might as well pass them through, as they are part of a complete GenerateContentRequest and may affect tokenCount in the future.

### DIFF
--- a/packages/main/src/requests/request-helpers.test.ts
+++ b/packages/main/src/requests/request-helpers.test.ts
@@ -17,10 +17,15 @@
 
 import { expect, use } from "chai";
 import * as sinonChai from "sinon-chai";
+import chaiDeepEqualIgnoreUndefined from "chai-deep-equal-ignore-undefined";
 import { Content } from "../../types";
-import { formatGenerateContentInput } from "./request-helpers";
+import {
+  formatCountTokensInput,
+  formatGenerateContentInput,
+} from "./request-helpers";
 
 use(sinonChai);
+use(chaiDeepEqualIgnoreUndefined);
 
 describe("request formatting methods", () => {
   describe("formatGenerateContentInput", () => {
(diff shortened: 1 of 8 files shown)


packages/main/src/requests/request-helpers.ts
```diff
@@ -111,9 +113,30 @@ function assignRoleToPartsAndValidateSendMessageRequest(
...
+  //     model: modelParams?.model,
+  //     generationConfig: modelParams?.generationConfig,
+  //     safetySettings: modelParams?.safetySettings,
+  //     tools: modelParams?.tools,
+  //     toolConfig: modelParams?.toolConfig,
+  //     systemInstruction: modelParams?.systemInstruction,
+  //     cachedContent: modelParams?.cachedContent?.name,
+  //     contents: [],
+  //   },
+  // };
```
DellaBitta, line 139: Should this be removed?

packages/main/src/requests/request-helpers.test.ts
```diff
@@ -172,4 +177,104 @@ describe("request formatting methods", () => {
       });
     });
   });
+  describe("formatCountTokensInput", () => {
+    it("formats a text string into a count request", () => {
+      const result = formatCountTokensInput("some text content", {
+        model: "gemini-1.5-flash",
+      });
+      // expect(result.generateContentRequest.model).to.equal("gemini-1.5-flash");
+      // expect(result.generateContentRequest.contents).to.equal("gemini-1.5-flash");
```
DellaBitta, line 186: Remove these?

DellaBitta (changes requested):

packages/main/src/requests/request-helpers.ts
hsubox76: Yep, thanks!

packages/main/src/requests/request-helpers.test.ts
hsubox76: Done.

DellaBitta (approved):

PR #226, opened 2024-08-14
Remove monorepo structure and tooling

hsubox76:
When this repo was originally created, we expected to have to create 2 npm packages for web and node. This never materialized so I am removing the unnecessary monorepo structure and tooling. This puts the CHANGELOG at the top level and removes the duplicate READMES, making the repo less complicated to read.

- Move contents of `packages/main` to top level.
- Use top level package.json as json of the actual published package instead of maintaining an umbrella package.json that does not represent a real NPM package
- Remove lerna and its config
- Update paths


package.json
```diff
@@ -1,16 +1,47 @@
...
-    "format": "yarn ts-node scripts/run-format.ts",
-    "format:check": "yarn ts-node scripts/check-format.ts"
+    "build": "rollup -c && yarn api-report",
+    "test": "yarn lint && yarn test:node:unit",
+    "test:web:integration": "yarn build && yarn web-test-runner",
+    "test:node:unit": "TS_NODE_COMPILER_OPTIONS='{\"module\":\"commonjs\"}' mocha \"src/**/*.test.ts\"",
+    "test:node:integration": "yarn build && TS_NODE_COMPILER_OPTIONS='{\"module\":\"commonjs\"}' mocha \"test-integration/node/**/*.test.ts\"",
+    "lint": "eslint -c .eslintrc.js '**/*.ts' --ignore-path './.gitignore'",
+    "api-report": "api-extractor run -c api-extractor.json --local --verbose && api-extractor run -c api-extractor.server.json --local --verbose",
+    "docs": "yarn build && yarn api-documenter markdown -i ./temp/main -o ../../docs/reference/main && yarn api-documenter markdown -i ./temp/server -o ../../docs/reference/server",
```
dlarocque, line 42: I think we should replace `../../docs` with `./docs` here

samples/.eslintrc.js
```diff
@@ -0,0 +1,138 @@
...
+      { name: "xdescribe" },
+      { name: "parseInt", message: "tsstyle#type-coercion" },
+      { name: "parseFloat", message: "tsstyle#type-coercion" },
+    ],
+    "no-array-constructor": "error",
+    "sort-imports": [
+      "error",
+      {
+        ignoreCase: false,
+        ignoreDeclarationSort: true, // don"t want to sort import lines, use eslint-plugin-import instead
```
dlarocque, line 122: Typo: don"t

scripts/check-format.ts
```diff
@@ -48,4 +49,7 @@ async function checkFormat() {
   process.exit(exitCode);
 }
 
-checkFormat();
+checkFormat().catch((e) => {
+  console.error(e);
+  process.exit(1);
```
dlarocque, line 54: Question: Is the only difference between this and just calling `checkFormat()` that this won't throw and unhandled promise rejection? Or are there other reasons for not just letting the error bubble up

dlarocque (changes requested):

scripts/check-format.ts
hsubox76: Functionally I'm not sure what the difference is - I think it will exit normally with code 1 instead of an "unhandled" error. Practically the difference is that if I don't put a catch on an async function call, the linter will yell at me.

DellaBitta (dismissed):
Nice finds, @dlarocque .

DellaBitta:
Complexity reduction, nice work! 👍

samples/.eslintrc.js
hsubox76: Fixed both instances

dlarocque (approved):

PR #235, opened 2024-08-22
Stop hardcoding "python" in text convenience method that returns code execution markdown

hsubox76:
Python is currently the only language the code execution feature will return but we still shouldn't hardcode it. The language is a response property for a reason.


DellaBitta (approved):

andrewheard (approved):
Thanks, Christina! LGTM.

PR #238, opened 2024-08-24
Fixed RequestUrl, it now adds API Key to the URL

Albert-Gao:
Without this fix, it throws such error when invoking:

> [GoogleGenerativeAI Error]: Error fetching from https://generativelanguage.googleapis.com/v1beta/models/gemini-1.5-flash:generateContent: [403 ] Method doesn't allow unregistered callers (callers without established identity). Please use API Key or other form of API consumer identity to call this API.

image

After applying the fix, it works now.


Albert-Gao:
I have submitted my CLA.

Albert-Gao:
When will this PR get reviewed.....still blocking....

PR #252, opened 2024-09-10
Support file uploads using a stream

mike-marcacci:
The current implementation requires a file to exist in the local filesystem, which is an uncommon scenario for production use cases. Instead, files are typically either stored in an appropriate external storage system like GCS or are supplied directly by the user over the network.

This change adds the ability to supply a file upload as a stream, and improves the existing filesystem flow by streaming the file's contents rather than attempting to load it into memory first (it was previously using the blocking call, which is only really appropriate for init stage file loading, quick demos, or systems that don't need to support concurrent requests).

### DIFF
--- a/src/server/request.ts
+++ b/src/server/request.ts
@@ -110,7 +110,7 @@ export async function makeServerRequest(
 
   if (typeof body === "string" || body instanceof Blob) {
     requestInit.body = body;
-  } else if (body[Symbol.asyncIterator]) {
+  } else if (body?.[Symbol.asyncIterator]) {
     // Note that in later versions, the signature `fetch` is updated to accept any AsyncIterator,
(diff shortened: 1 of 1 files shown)


mike-marcacci:
Heyo - I just want to check whether a streaming approach is something that will be considered in this project. We've been patching this locally during development, but if this is unlikely to be merged or made possible through an alternate implementation, we'll just go ahead and publish a fork of this package.

mike-marcacci:
In case it's helpful to anyone, [here](https://gist.github.com/mike-marcacci/0d22a7a04acbcb517f9b6c117b35f1ac)'s a pretty straightforward workaround, largely just copying the existing implementation.

CamdenClark:
Really appreciate you trying to PR this, and thank you for the workaround.

PR #254, opened 2024-09-10
Add ability to set modelParams on getGenerativeModelFromCachedContent()

hsubox76:
Add a modelParams arg to getGenerativeModelFromCachedContent(), allowing users to set generationConfig and safetySettings.

### DIFF
--- a/common/api-review/generative-ai.api.md
+++ b/common/api-review/generative-ai.api.md
@@ -481,7 +481,7 @@ export class GoogleGenerativeAI {
     // (undocumented)
     apiKey: string;
     getGenerativeModel(modelParams: ModelParams, requestOptions?: RequestOptions): GenerativeModel;
-    getGenerativeModelFromCachedContent(cachedContent: CachedContent, requestOptions?: RequestOptions): GenerativeModel;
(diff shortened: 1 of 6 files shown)


common/api-review/generative-ai.api.md
```diff
@@ -481,7 +481,7 @@ export class GoogleGenerativeAI {
     // (undocumented)
     apiKey: string;
     getGenerativeModel(modelParams: ModelParams, requestOptions?: RequestOptions): GenerativeModel;
-    getGenerativeModelFromCachedContent(cachedContent: CachedContent, requestOptions?: RequestOptions): GenerativeModel;
+    getGenerativeModelFromCachedContent(cachedContent: CachedContent, modelParams?: Partial<ModelParams>, requestOptions?: RequestOptions): GenerativeModel;
```
dlarocque, line 484: Why do we make this `Partial<>`? Is `ModelParams.model` only optional when it's passed to this method?

dlarocque (commented):
Is the failed node integration test a flake?

common/api-review/generative-ai.api.md
hsubox76: None of the params are required here, and the whole object is optional, because `cachedContent` should have everything it needs to initialize (it always has "model").

PR #266, opened 2024-09-26
Add responseLogProbs and logProbs parameters to generateContentReq

shilpakancharla:
Following up on b/369367297.

New input parameter are listed here GenerateContentRequest:

* optional bool response_logprobs
* optional int32 logprobs

New output parameter are listed here GenerateContentResponse:

* double avg_logprobs
* LogprobsResult logprobs_result


.changeset/funny-pillows-whisper.md
hsubox76, line 1: Did you run yarn changeset twice? Looks like there's 2 files. You can just delete this one, looks like the other one is more complete. For future reference, if you want to update the text in a changeset, you can just do it in the file directly, no need to run yarn changeset again.

types/responses.ts
```diff
@@ -87,6 +87,41 @@ export interface GenerateContentResponse {
...
+   * The chosen candidates may or may not be in topCandidates.
+   */
+  chosenCandidates: Candidate[];
+}
+
+/**
+ * Candidate for the logprobs token and score.
+ * @public
+ */
+export interface Candidate {
```
hsubox76, line 108: Can we name this more specifically? We already have `GenerateContentCandidate` which is the most common type of `candidate` most users will interact with. Maybe `LogProbsCandidate`?

types/responses.ts
```diff
@@ -124,6 +159,10 @@ export interface GenerateContentCandidate {
   finishMessage?: string;
   safetyRatings?: SafetyRating[];
   citationMetadata?: CitationMetadata;
+  /** Average log probability score of the candidate. */
+  avgLogProbs?: number;
```
hsubox76, line 163: since it's `avg_logprobs` in the proto, the capitalization should translate to `avgLogprobs` here.

types/responses.ts
```diff
@@ -124,6 +159,10 @@ export interface GenerateContentCandidate {
...
+  /** Log-likelihood scores for the response tokens and top tokens. */
+  logProbsResult?: LogprobsResult;
```
hsubox76, line 165: Since the proto has `logprobs_result`, this would translate to `logprobsResult` in camelcase. This should be the capitalization in the response, you can double check with curl I think.

types/responses.ts
hsubox76, line 163: A good place to test responses might be in this file: https://github.com/google-gemini/generative-ai-js/blob/main/src/methods/generate-content.test.ts#L86 That's an example of testing citations in the response, which might be the closest thing to this feature. You will need to create a json file that represents a fake response from the backend. In that citation test example, the corresponding file is here: https://github.com/google-gemini/generative-ai-js/blob/0d4fe068cf8deb30ddc14bc3d33446696219bea5/test-utils/mock-responses/unary-success-citations.json You can create it yourself, or if you want to be safe, curl the backend and use a real response. Watch out for carriage returns and hidden whitespace characters if you copy paste from a real response.

common/api-review/generative-ai.api.md
```diff
@@ -9,7 +9,9 @@ export interface BaseParams {
     frequencyPenalty?: number;
     // (undocumented)
     generationConfig?: GenerationConfig;
+    logProbs?: number;
     presencePenalty?: number;
+    responseLogProbs?: boolean;
```
hsubox76, line 14: response_logprobs => responseLogprobs

common/api-review/generative-ai.api.md
```diff
@@ -9,7 +9,9 @@ export interface BaseParams {
     frequencyPenalty?: number;
     // (undocumented)
     generationConfig?: GenerationConfig;
+    logProbs?: number;
```
hsubox76, line 12: logprobs (no capital P)

.changeset/funny-pillows-whisper.md
hsubox76: I still see the extra file?

types/requests.ts
```diff
@@ -35,12 +35,21 @@ export interface BaseParams {
...
   /**
    * Frequency penalty applied to the next token's logprobs, multiplied by the
    * number of times each token has been seen in the respponse so far.
    */
-  frequencyPenalty?: number
+  frequencyPenalty?: number;
+  /**
+   * If True, export the logprobs results in response.
+   */
+  responseLogProbs?: boolean;
```
hsubox76, line 47: I'm sorry, I just looked at the proto and saw that these go inside GenerationConfig? That's down in line 104. It seems the same is the case for presencePenalty and frequencyPenalty as well.

src/models/generative-model.test.ts
```diff
@@ -120,9 +123,7 @@ describe("GenerativeModel", () => {
           value.includes("be friendly") &&
           value.includes("temperature") &&
           value.includes("testField") &&
-          value.includes(HarmBlockThreshold.BLOCK_LOW_AND_ABOVE) &&
-          value.includes("presencePenalty") &&
-          value.includes("frequencyPenalty")
```
hsubox76, line 125: These can be left in, and the two new ones added as well. The first set of checks in 98-101 only check that these properties were added to the model class. This set of checks is even more important as it ensures that when generateContent is called, these are actually passed in the body to makeModelRequest. This just searches the body text for any mention of the param, so it doesn't matter if it's nested - it's all just one big JSON string.

src/models/generative-model.test.ts
```diff
@@ -236,9 +235,7 @@ describe("GenerativeModel", () => {
           value.includes("topK") &&
           value.includes("newTestField") &&
           !value.includes("testField") &&
-          value.includes(HarmCategory.HARM_CATEGORY_HARASSMENT) &&
-          value.includes("presencePenalty") &&
-          value.includes("frequencyPenalty")
```
hsubox76, line 241: Same comment as above, I would add these back in.

hsubox76 (approved):

PR #276, opened 2024-10-09
Imagen

junyanxu:
Doc change not applied yet. Please focus on the main logic.


types/requests.ts
```diff
@@ -245,3 +245,78 @@ export interface CodeExecutionTool {
...
+   */
+  numberOfImages?: number;
+  /**
+   * Width of the image. One of the Width/Height sizes must be 256 or 1024.
+   */
+  width?: number;
+  /**
+   * Height of the image. One of the Width/Height sizes must be 256 or 1024.
+   */
+  height?: number;
```
dlarocque, line 273: Do we provide any validation for these values? What kind of error will the user see if they provide an invalid value (e.g. numerOfImages = 10)?

types/generated-media.ts
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * Generated video.
+ * @public
+ */
+export interface GeneratedVideo {
+  /**
+   * Video bytes.
+   */
+  video?: string;
```
dlarocque, line 36: Is the `video` and `imageBytes` that we get back encoded as base64?

types/requests.ts
junyanxu: The value validation would be done at the backend server. If they provide the wrong argument, the response status would be INVALID_ARGUMENT.

types/requests.ts
```diff
@@ -245,3 +245,78 @@ export interface CodeExecutionTool {
...
+  /**
+   * Changes the aspect ratio of the generated image Supported
```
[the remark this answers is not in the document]
junyanxu: Good call.

package.json
```diff
@@ -88,5 +89,9 @@
   "bugs": {
     "url": "https://github.com/google/generative-ai-js/issues"
   },
-  "homepage": "https://github.com/google/generative-ai-js#readme"
+  "homepage": "https://github.com/google/generative-ai-js#readme",
+  "dependencies": {
+    "@types/google-protobuf": "^3.15.12",
+    "google-protobuf": "^3.21.4"
```
hsubox76, line 95: Where are these used?

package.json
```diff
@@ -46,6 +46,7 @@
   "devDependencies": {
     "@changesets/cli": "^2.27.1",
     "@esm-bundle/chai": "^4.3.4-fix.0",
+    "@google-cloud/asset": "^5.7.0",
```
hsubox76, line 49: Where is this used?

types/predict.ts
```diff
@@ -0,0 +1,48 @@
+export type PredictServiceBasicValueType =
```
hsubox76, line 1: I noticed none of the new types that are added seem to be exported from `types/index.ts` which means the user won't have access to them. I assume you want the developer to have access to some of them, can you make sure they're exported from `types/index.ts`?

src/requests/request-helpers.ts
```diff
@@ -177,3 +178,40 @@ export function formatEmbedContentInput(
...
+}
+
+export function convertFromImageGenerationRequest(
+  model: string,
+  request: ImageGenerationRequest,
+): PredictRequest {
+  const instances = [{ prompt: request.prompt }];
+  const sampleImageSize = Math.max(request.width || 0, request.height || 0);
+  const parameters = {
+    negativePrompt: request.negativePrompt,
```
hsubox76, line 201: Can you start with `...request` for all the request parameters that don't need any modification and then add the modified ones after?

types/requests.ts
```diff
@@ -245,3 +245,81 @@ export interface CodeExecutionTool {
...
+  /**
+   * Text prompt for the image.
+   */
+  prompt: string;
+  /**
+   * A description of what you want to omit in the generated images.
+   */
+  negativePrompt?: string;
+  /**
+   * Number of images to generate. Range: 1..8.
```
dlarocque, line 263: When I tried passing `numberOfImages: 8`, I got an error saying 4 is the max.

```
GoogleGenerativeAIFetchError: [GoogleGenerativeAI Error]: Error fetching from https://generativelanguage.googleapis.com/v1beta/models/imagen-3.0-generate-001:predict: [400 Bad Request] Image generation failed with the following error: Invalid sample count 8 in request. Max number of samples prebuilt model can return is 4. Min number is 1
```

types/requests.ts
```diff
@@ -245,3 +245,81 @@ export interface CodeExecutionTool {
...
+  aspectRatio?: "1:1" | "9:16" | "16:9" | "4:3" | "3:4";
+  /**
+   * Controls the strength of the prompt. Suggested values are:
+   * * 0-9 (low strength) 
+   * * 10-20 (medium strength) 
+   * * 21+ (high strength)
+   */
+  guidanceScale?: number;
+  /**
+   * hich image format should the output be saved as.
```
dlarocque, line 292: Typo: hich -> Which

types/requests.ts
```diff
@@ -245,3 +245,81 @@ export interface CodeExecutionTool {
...
+   */
+  numberOfImages?: number;
+  /**
+   * Width of the image. One of the Width/Height sizes must be 256 or 1024.
+   */
+  width?: number;
+  /**
+   * Height of the image. One of the Width/Height sizes must be 256 or 1024.
+   */
+  height?: number;
```
dlarocque, line 273: When I specify `width` and `height` to `256`, I get back an image that is `1024x1024`. I'm not sure whether this is a backend issue or not.

src/requests/request-helpers.ts
```diff
@@ -177,3 +178,40 @@ export function formatEmbedContentInput(
...
+  }
+  return formattedRequest;
+}
+
+export function convertFromImageGenerationRequest(
+  model: string,
+  request: ImageGenerationRequest,
+): PredictRequest {
+  const instances = [{ prompt: request.prompt }];
+  const sampleImageSize = Math.max(request.width || 0, request.height || 0);
```
dlarocque, line 199: If we use the max of `width` and `height` as the image size and only pass that one dimension, why not just have one `imageSize` option?

junyanxu:
Sorry guys, I messes up the git after doing a merge. I open up a new branch for us to review.

PR #285, opened 2024-10-18
Pull Imagen feature

junyanxu:
I messed up the old branch after doing a merge. Create this new branch for us to keep reviewing.

I have resolved all comments except one from @hsubox76 

The ...request will cause testing error like this, because original request has more parameters than `parameters` could accept. And in new predict api the `parameters` are not typed.  

```shell

      AssertionError: expected { model: 'test_model', …(2) } to deeply equal { …(3) }
      + expected - actual

         "parameters": {
           "aspectRatio": "1:1"
           "compressionQuality": 12
           "guidanceScale": 5
      -    "height": 1024
           "language": "ko"
           "negativePrompt": "Not a dog"
      -    "numberOfImages": 1
           "outputMimeType": "image/png"
      -    "prompt": "Create a cat"
           "safetyFilterLevel": "block_medium_and_above"
           "sampleCount": 1
           "sampleImageSize": 1024
      -    "width": 512
         }
       }
      
```

PR #287, opened 2024-10-31
Add watermark support for imagen

junyanxu:
Add watermark support to imagen

Example curl:

curl https://generativelanguage.googleapis.com/v1beta/models/image-verification-001:predict?key=$API_KEY     -H 'Content-Type: application/json'     -X POST     -d '{ "instances":[
      { "image": {"bytesBase64Encoded": <RAW_IMAGE_BYTES>}}
    ],
    "parameters": {"watermarkVerification": true}
}'


tonycoco:
Reference to: #276 #285 are we merging this soon?

PR #295, opened 2024-11-12
Allow Schema types to be written as inline string literals.

rictic:
Declaring SchemaType as an enum means that users must reference the schema type off the enum, like SchemaType.OBJECT. That's useful, but it's also nice to be able to just write 'object', and with TypeScript that's equally type safe, IDEs are equally helpful with autocomplete, etc!

This is a backwards compatible change that allows users to write it either way.


junyanxu (approved):
Approved after offline discussion

junyanxu (approved):

rictic:
I rebased and regenerated the docs

PR #298, opened 2024-11-16
Feat/add buffer support

abdelkd:
This Pull-Request adds support for using `Buffer` inside of `uploadFile` function instead of having file to be on local disk. It can be used for example to process a form request and upload immediately instead of saving it on local and then uploading it.

### DIFF
--- a/.changeset/fuzzy-cows-approve.md
+++ b/.changeset/fuzzy-cows-approve.md
@@ -1,0 +1,5 @@
+---
+"@google/generative-ai": minor
+---
+
+added support for buffers in uploadFile
--- a/common/api-review/generative-ai-server.api.md
+++ b/common/api-review/generative-ai-server.api.md
@@ -4,6 +4,8 @@
 
 ```ts
 
+/// <reference types="node" />
+
 // @public
 export interface CachedContent extends CachedContentBase {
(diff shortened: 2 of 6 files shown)


jimmywarting:
Thank you for this, this is the kind of smart thinking i like the most! 
A package should assume as little about the system it runs on as much as possible. 
I sometimes think it's bad that some application assumes it have access to some basic I/O operations when it could in fact run on multiple possible environment, like NodeJS, Bun, Deno, Browsers, Web Workers etc, it can be so sandboxed as much as possible. 

a package should basically work as a stdin and stdout only and learn from the [onion architecture](
https://web.archive.org/web/20190515082139/https://www.codeguru.com/csharp/csharp/cs_misc/designtechniques/understanding-onion-architecture.html)

One small flaw in this PR is that you used `node:Buffer` instead of plain `Uint8Array` or `ArrayBuffer` instead. I'm not a huge fan of NodeJS specific things. it pollutes the env in other runtimes.

hkt74 (commented):
Thanks for making the contribution, could you regenerate the api docs to resolve the conflicts?

hkt74:
I cloned this PR to #365 which resolved the merge conflict

PR #318, opened 2025-01-01
Adding search tool demo

Leask:


Leask:
Having signed the CLA, what are the next steps?

samples/search_tool.js
```diff
@@ -0,0 +1,50 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {
+  GoogleGenerativeAI,
+} from "@google/generative-ai";
+
+async function searchGrounding() {
```
hkt74, line 22: what do you think to rename this to google_search and move the function to samples/search_grounding.js

so we can keep the similar test cases together and maybe worth to add a comment that 

* 'googleSearchRetrieval' supports gemini 1.5 models
* 'googleSearch' support gemini 2.0 models

hkt74, line 22: what do you think to rename this to google_search and move the function to samples/search_grounding.js

so we can keep the similar test cases together and maybe worth to add a comment that

'googleSearchRetrieval' is available in gemini 1.5 models
'googleSearch' is available in gemini 2.0 models

Leask: Accepted via: https://github.com/google-gemini/generative-ai-js/pull/318/commits/5d7e12438d80e5c02ba8768fe874a2ec9a70e9b7 
Thanks! 🤝

hkt74 (approved):

Leask:
Thanks! 😄

PR #357, opened 2025-02-27
Introduce a new error type to represent AbortError

### LINKED ISSUE #303: Unhandled `AbortError` causes `Unhandled Rejection`
### Description of the bug:

When an `AbortController` is triggered (e.g., via timeout or manual abort), the SDK throws an `AbortError` that **cannot be caught using `try...catch`**. This results in an `Unhandled Rejection`, forcing developers to handle it globally with `process.on('unhandledRejection', ...)`.

Here is the error log:

```
[Unhandled rejection]
DOMException [AbortError]: This operation was aborted
    at new DOMException (node:internal/per_context/domexception:53:5)
    at AbortController.abort (node:internal/abort_controller:401:18)
    at EventTarget.<anonymous> (O:\work_space\github.com\@lenML\project1\node_modules\.pnpm\@google+generative-ai@0.21.0\node_modules\@google\generative-ai\dist\index.js:430:28)         
    at [nodejs.internal.kHybridDispatch] (node:internal/event_target:786:20)
    at EventTarget.dispatchEvent (node:internal/event_target:721:26)
    at abortSignal (node:internal/abort_controller:369:10)
    at AbortController.abort (node:internal/abort_controller:403:5)
```

Suspected problematic code: 
https://github.com/google-gemini/generative-ai-js/blob/2df2af03bb07dcda23b07af1a7135a8b461ae64e/src/requests/stream-reader.ts#L78-L128

### Actual vs expected behavior:

- **Actual:**  
  `AbortError` is not catchable by `try...catch`, causing unhandled promise rejections.

- **Expected:**  
  `AbortError` should be catchable within `try...catch` in user code, without requiring global error handlers.

none

hkt74:
Fixes #303


src/requests/request.ts
```diff
@@ -172,7 +173,12 @@ export async function makeRequest(
 
 function handleResponseError(e: Error, url: string): void {
   let err = e;
-  if (
+  if (err.name === "AbortError") {
```
IvanLH, line 176: I don't think this is the fix, on this path all errors are already wrapped into GoogleGenerativeAI Error, the issue is that on StreamReader we are not catching errors when the stream is read.

IvanLH (changes requested):

src/requests/request.ts
IvanLH: Do we need this part changed?

src/requests/stream-reader.ts
```diff
@@ -89,38 +92,54 @@ export function getResponseStream<T>(
...
+            let err = e;
+            err.stack = e.stack;
+            if (err.name === "AbortError") {
+              err = new GoogleGenerativeAIAbortError(
+                "Request aborted when reading from the stream",
+              );
+            } else {
+              err = new GoogleGenerativeAIError(
+                "Error reading from the stream",
               );
```
IvanLH, line 139: Could we just do  throw new GoogleGenerativeAIError('Failed to perform operation', { cause: error }); Error supports cause, and I think it negates the need for specific classes.

IvanLH (approved):

KiGamji:
@danny-avila

Issue #188, opened 2024-06-27
ResponseSchema with enum gives 400 Bad Request Error


navarrodiego:
### Description of the bug:

I am using the responseSchema functionality of Gemini. I can't make it work with enums.

This is how i initialize the model:

```
new GenerativeModel(
        geminiApiKey, {
        model: "gemini-1.5-pro-latest",
        generationConfig: {
            responseMimeType: 'application/json',
            responseSchema: schema
        }
    })
```

This is a test response schema **that works:**

```
export const schema: ResponseSchema = {
  description: "Complete fields with information of the product",
  type: FunctionDeclarationSchemaType.ARRAY,
  items: {
    type: FunctionDeclarationSchemaType.OBJECT,
    properties: {
      typeOfSleeve: {
        type: FunctionDeclarationSchemaType.STRING,
        description: "",
        nullable: true
      },
    },
    required: [
     "typeOfSleeve"
    ]
  }
};
```

The exact same schema, with an enum fails:
```
export const schema: ResponseSchema = {
  description: "Complete fields with information of the product",
  type: FunctionDeclarationSchemaType.ARRAY,
  items: {
    type: FunctionDeclarationSchemaType.OBJECT,
    properties: {
      typeOfSleeve: {
        type: FunctionDeclarationSchemaType.STRING,
        description: "",
        nullable: true,
        enum: [
          "no sleeves",
          "short",
          "3/4",
          "long"
        ]
      },
    },
    required: [
     "typeOfSleeve"
    ]
  }
};
```
It gives me the error: 
`[GoogleGenerativeAI Error]: Error fetching from https://generativelanguage.googleapis.com/v1beta/models/gemini-1.5-pro-latest:generateContent: [400 Bad Request] Request contains an invalid argument.`

My prompt is:
```
const prompt = (productInformation: string): string => {
  return `Follow this JSON: <JSONSchema>${JSON.stringify(
    schema
  )}</JSONSchema>

Product information: ${productInformation}
`;
}
```

### Actual vs expected behavior:

Actual: 400 Bad Request Error
Expected: Correct 200 status code and response

### Any other information you'd like to share?

I also think that the ResponseSchema interface of TypeScript has errors.

nicolab28:
I've had the same error since around Wednesday.
I didn't change the code, and until Tuesday it worked perfectly.

As soon as there's an enum in a tool, I get this error:
[...]

PR #369, opened 2025-03-04
Adds types which will enforce format rules when using String Schemas

IvanLH:
This change enforces the following rules, which fix Issue #188

```
// Allowed
const schema: Schema = {
    type: SchemaType.String
}

// Allowed
const schema: Schema = {
    type: SchemaType.String,
    format: "date-time"
}

// Allowed
const schema: Schema = {
    type: SchemaType.String,
    format: "enum",
    enum: ["val1", "val2"],
}

// TS Error
const schema: Schema = {
    type: SchemaType.String,
    format: "enum",
}

// TS Error
const schema: Schema = {
    type: SchemaType.String,
    format: "date-time",
    enum: ["val1", "val2"]
}

// TS Error
const schema: Schema = {
    type: SchemaType.String,
    enum: ["val1", "val2"]
}


// TS Error
const schema: Schema = {
    type: SchemaType.String,
    format: "bytes"
}


```


types/function-calling.ts
```diff
@@ -162,15 +162,42 @@ export interface IntegerSchema extends BaseSchema {
...
-  enum?: string[];
-  // Note that the API accepts the `pattern`, `minLength`, and `maxLength`
-  // fields, but they may only be advisory.
-  // The `format` field is not (at time of writing) supported on strings.
+
+  // Note: These undefined values are needed to help the type system, they won't,
+  // be passed to the API as they are undefined
+  format?: "date-time" | undefined;
+
+  enum?: never;
```
hkt74, line 182: is this needed?

IvanLH:
> are there tests cover enum schema?

I thought about it, however the failures are at compilation time. Technically the code is valid, just the TS compiler won't let you unless you annotate with an ignore tag.

hkt74 (approved):

Issue #54, opened 2024-02-27
Add support for list models


sebastianbenz:
It'd be great if you could add support for the `list_models` API https://ai.google.dev/api/python/google/generativeai/list_models.

shatfield4:
Are there any plans to have this function implemented soon?

BobDickinson:
Still an issue in the new SDK: https://github.com/googleapis/js-genai/issues/473

If you're following, please go give that an upvote.

PR #374, opened 2025-03-06
feat: Add support for list models

krishnaagrawal7508:
This PR attempts to fix https://github.com/google-gemini/generative-ai-js/issues/54

### DIFF
--- a/common/api-review/generative-ai-server.api.md
+++ b/common/api-review/generative-ai-server.api.md
@@ -4,6 +4,8 @@
 
 ```ts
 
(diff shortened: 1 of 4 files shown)


krishnaagrawal7508:
@IvanLH @hkt74, since `listModels()` retrieves all models in the system, I believe placing it in `GoogleGenerativeAI` makes sense as it is not a model specific function. However, the overall structure isn't clearly documented. I made a PR with the above stated placement, or do you have a preferred hierarchy in mind?

PR #385, opened 2025-03-07
Enabled tuned models listing with custom request function

akp111:
I have added the functionality to enable list_tuned_models in js.
For this, I had to make some changes in the request.ts to accept get with  query parameters and also to stop accepting model as a mandatory parameter


BobDickinson:
+1 on this.  Any ETA?  I only ask because it's been pending review for over a month.  My app just has a hardcoded list until this is released, which is not ideal.

PR #401, opened 2025-03-08
Enhance generateContentStream with streamCallbacks support. Fixes #322

### LINKED ISSUE #322: Add support for callbacks on streamed response
### Description of the feature request:

Add callbacks on streamed response to allow developers to easily manage generating streamed chat responses with callbacks such as saving the prompt to database or any server side call.

### What problem are you trying to solve with this feature?

This feature is helpful for a good DX. Currently the user will need to manage the response manually. For example I had to implement this function to get a streamed response with callbacks:
```javascript
export const getStreamedAIResponse = async ({ prompt, onData, onEnd }) => {
  const response = await model.generateContentStream([prompt]);

  let textBuf = '';

  while (true) {
    const { done, value } = await response.stream.next()
    if (done) {
      await onEnd(textBuf);
      return
    }

    if (value) {
      const currentText = value.text()
      onData(value.text())
      textBuf += currentText
    }
  }
}
```

Initially what this function does is allowing the user to sign a `onData` and  `onEnd` functions and the server will call one on every chunk the LLM API responds and the other at the end of generating the response. This way It will give the user a Chat-like streamed response without the need to way for the whole response and also allows the developer to save the response on the end of the generation without relying on any client-side callback.

It would be good to expose a function like this: 
```javascript
const response = generateStreamedResponse([prompt, {fileData: {fileUri, mimeType}}], {
  onData: (chunk) => {},
  onEnd: (repsonse) => {},
  onError: (error) => {}
})
```

The returned response is of type `Response` which would allow for directly returning the value in the `Route handlers` in Next.js or the `API endpoint` in SvelteKit.

This could also be beneficial in Express.js and Hono

cestercian:
- Adds `streamCallbacks` as an optional argument to `generateContentStream`.  
- Enables users to receive text responses via callbacks without handling Node.js streams.  
- Includes test cases for the new functionality.  

#### Example Usage:
```ts
await model.generateContentStream("What is 2 + 2", {}, {
   onData: (chunk: string) => console.log(chunk),
   onDone: (fullText: string) => console.log(fullText),
});

### DIFF
--- a/common/api-review/generative-ai.api.md
+++ b/common/api-review/generative-ai.api.md
@@ -525,7 +525,7 @@ export class GenerativeModel {
     countTokens(request: CountTokensRequest | string | Array<string | Part>, requestOptions?: SingleRequestOptions): Promise<CountTokensResponse>;
     embedContent(request: EmbedContentRequest | string | Array<string | Part>, requestOptions?: SingleRequestOptions): Promise<EmbedContentResponse>;
     generateContent(request: GenerateContentRequest | string | Array<string | Part>, requestOptions?: SingleRequestOptions): Promise<GenerateContentResult>;
-    generateContentStream(request: GenerateContentRequest | string | Array<string | Part>, requestOptions?: SingleRequestOptions): Promise<GenerateContentStreamResult>;
+    generateContentStream(request: GenerateContentRequest | string | Array<string | Part>, requestOptions?: SingleRequestOptions, callbacks?: StreamCallbacks): Promise<GenerateContentStreamResult>;
     // (undocumented)
     generationConfig: GenerationConfig;
     // (undocumented)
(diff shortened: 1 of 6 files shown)


demoncoder-crypto:
I am working on this issue let me know if its up to expectation

avinesh2101 (commented):
Hey! This change has already been proposed in [[Original PR #371](https://github.com/google-gemini/generative-ai-js/pull/371/files)]. Submitting duplicate PRs doesn’t add value and can create unnecessary noise for maintainers. It’s best to check existing PRs before submitting new ones. Maybe consider contributing in other meaningful ways!

cestercian:
Hey @avinesh2101 .Thanks for pointing that out. I didn’t realize this change had already been proposed in PR #371. I’ll make sure to check existing PRs more carefully next time. Appreciate the guidance! I’d love to contribute in other ways—if there are any areas that need help, I’d be happy to take a look.

PR #420, opened 2025-03-13
fix: After an API error, all subsequent calls fail with the same error

### LINKED ISSUE #417: After an API error, all subsequent calls fail with the same error
#### **`@google/generative-ai` version:** 0.24.0

After any error occurs in the API (e.g., 404 Not Found, 429 Too Many Requests, or 503 Service Unavailable), all subsequent API calls return the same error, even if the underlying issue is resolved.

When calling `sendMessage()`, if an error occurs, all subsequent calls to `sendMessage()` and `getHistory()` on the same `chatSession` instance instantly throw the same error.

This happens because `_sendPromise` remains in a rejected state and is not reset after the failure, causing all future requests to fail immediately with the same error, even if the root cause has been fixed.

#### Actual
- After an API call fails, all subsequent calls to `sendMessage()` and `getHistory()` on the same `chatSession` instance instantly return the same error, even if the issue has been resolved.
- The error persists because `_sendPromise` remains in a rejected state and is reused for future requests.

#### Expected Behavior:
- If an API call fails (e.g., due to network issue, or API error), if the issue is fixed, should allow future API calls to succeed normally.
- Each request should be independent, meaning a past failure should not affect future calls.

### Code to Reproduce the Issue

Here’s a minimal example to reproduce the issue:

```js
import { GoogleGenerativeAI } from '@google/generative-ai'

const genAI = new GoogleGenerativeAI('...')
const model = genAI.getGenerativeModel({ model: 'gemini-1.5-pro' })
const chatSession = model.startChat()

await chatSession.sendMessage('Hello, how are you?') // OK, no error

// Now, let's force an error by setting a nonexistent model
// Note: This is just for demonstration purposes. This error can be caused by other reasons,
// such as network issues, 429 too many requests, 503 service unavailable, etc.

// Force an error from the Gemini API by setting a nonexistent model
chatSession.model = 'models/some-nonexistent-model'

// This will throw an error (expected)
await chatSession.sendMessage('Hello, how are you?').catch((error) => {
  console.error('>> ERROR (expected):', error.message, '\n')
  // ERROR: /v1beta/models/some-other-model:generateContent: [404 Not Found]
})

// Fix the model; it shouldn't give any more problems from here
chatSession.model = 'models/gemini-1.5-pro'

await chatSession.sendMessage('Hello, how are you?').catch((error) => {
  console.error('>> UNEXPECTED ERROR:', error.message, '\n')
  // Keep getting the same error, and the old error is still there
  // ERROR: /v1beta/models/some-other-model:generateContent: [404 Not Found]
  //                ^^^^^^^^^^^^^^^^
})

// Other methods will also throw the same error, like getHistory
await chatSession.getHistory().catch((error) => {
  console.log('>> UNEXPECTED ERROR FETCHING HISTORY:', error.message, '\n')
  // ERROR: /v1beta/models/some-other-model:generateContent: [404 Not Found]
  //                ^^^^^^^^^^^^^^^^
})

// This is because the _sendPromise is still rejected, and is awaiting on each call
// console.log(chatSession._sendPromise) // Promise { <rejected> }

// We need to reset it to a resolved promise
chatSession._sendPromise = Promise.resolve()

await chatSession.sendMessage('Hello, how are you?') // OK, no error

console.log('>> FIXED')
```

josejefferson:
Fix #417

### DIFF
--- a/src/methods/chat-session.ts
+++ b/src/methods/chat-session.ts
@@ -126,14 +126,13 @@ export class ChatSession {
           }
         }
         finalResult = result;
+      })
+      .catch((e) => {
+        // Resets _sendPromise to avoid subsequent calls failing and throw error.
+        this._sendPromise = Promise.resolve();
+        throw e;
       });
-
-    await this._sendPromise.catch((e) => {
-      // Resets _sendPromise and re-throws the error if it fails
-      this._sendPromise = Promise.resolve();
-      throw e;
-    });
-
+    await this._sendPromise;
     return finalResult;
   }
 
--- a/src/methods/chat-session.ts
+++ b/src/methods/chat-session.ts
@@ -128,12 +128,11 @@ export class ChatSession {
         finalResult = result;
       });
 
-    try {
-      await this._sendPromise;
-    } catch (error) {
+    await this._sendPromise.catch((e) => {
+      // Resets _sendPromise and re-throws the error if it fails
       this._sendPromise = Promise.resolve();
-      throw error;
-    }
+      throw e;
+    });
 
     return finalResult;
   }
--- a/src/methods/chat-session.ts
+++ b/src/methods/chat-session.ts
@@ -127,7 +127,14 @@ export class ChatSession {
         }
         finalResult = result;
       });
-    await this._sendPromise;
+    
+    try {
+      await this._sendPromise;
+    } catch (error) {
+      this._sendPromise = Promise.resolve();
+      throw error;
+    }
+
     return finalResult;
   }
 
--- a/src/methods/chat-session.ts
+++ b/src/methods/chat-session.ts
@@ -127,7 +127,7 @@ export class ChatSession {
         }
         finalResult = result;
       });
-    
+
     try {
       await this._sendPromise;
     } catch (error) {
--- a/src/methods/chat-session.test.ts
+++ b/src/methods/chat-session.test.ts
@@ -46,6 +46,27 @@ describe("ChatSession", () => {
         match.any,
       );
     });
+    it("generateContent errors should not persist", async () => {
+      const generateContentStub = stub(
+        generateContentMethods,
+        "generateContent",
+      ).rejects("generateContent failed");
+      const chatSession = new ChatSession("MY_API_KEY", "a-model");
+      await expect(chatSession.sendMessage("hello")).to.be.rejected;
+      expect(generateContentStub).to.be.calledWith(
+        "MY_API_KEY",
+        "a-model",
+        match.any,
+      );
+      restore();
+
+      // Subsequent call should succeed.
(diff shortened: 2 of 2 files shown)


src/methods/chat-session.ts
```diff
@@ -127,7 +127,14 @@ export class ChatSession {
         }
         finalResult = result;
       });
-    await this._sendPromise;
+
```
Annhiluc, line 130: Should we apply logic similar to https://github.com/google-gemini/generative-ai-js/pull/420/files#diff-250ed4bb03c096e3830327714961f4dbfde48df60b19c2918a2cdb4df06b6df1R203-R213?

As in, apply a .catch instead of a separate try/catch block here?

Annhiluc (changes requested):

src/methods/chat-session.ts
josejefferson: Do you mean something like this?

```js
await this._sendPromise.catch((error) => {
  this._sendPromise = Promise.resolve();
  throw error;
});
```

Annhiluc: Yes something like that. Could you check to see if this also works?

```
.then((result) => {
  // ...
  finalResult = result;
})
.catch((e) => {
  // Resets _sendPromise to avoid subsequent calls failing and throw error.
  this._sendPromise = Promise.resolve();
  throw e;
});
```

Annhiluc (changes requested):

src/methods/chat-session.ts
josejefferson: Yes, it works! I've already made the change.

Annhiluc (commented):
Thank you for making the changes! Could you also help add a unit test for this case in src/methods/chat-session.test.ts?

Something like:

```
it("generateContent errors should not persist", async () => {
      const generateContentStub = stub(
        generateContentMethods,
        "generateContent",
      ).rejects("generateContent failed");
      const chatSession = new ChatSession("MY_API_KEY", "a-model");
      await expect(chatSession.sendMessage("hello")).to.be.rejected;
      expect(generateContentStub).to.be.calledWith(
        "MY_API_KEY",
        "a-model",
        match.any,
      );

      // Subsequent call should succeed.
      const UpdatedGenerateContentStub = stub(
        generateContentMethods,
        "generateContent",
      );
      await expect(chatSession.sendMessage("hello")).to.not.be.rejected;
      expect(UpdatedGenerateContentStub).to.be.calledWith(
        "MY_API_KEY",
        "a-model",
        match.any,
      );
    });
```

josejefferson:
Sure! I've added a unit test for this case in `src/methods/chat-session.test.ts`. Let me know if you need any adjustments or additional test cases. Thanks for your help!

Annhiluc (approved):

PR #424, opened 2025-03-16
fix: Add missing fs import for file handling

SGNayak12:
Added **const fs = require("fs");**  to fix missing import.
Ensures proper file reading using **fs.readFileSync().**


hkt74 (commented):
what do you think to use ?

```
import * as fs from 'fs';
```

this repo uses ES modules

SGNayak12:
ok got it
**import * as fs from 'fs';**
This will import everything from the 'fs' module right..

hkt74 (approved):

PR #430, opened 2025-03-18
Add a GitHub Action to manage `status: awaiting user response`

amirh:
The idea is that any new comment on an issue with this label would get the label automatically removed.

I don't know how to test this without submitting, going to try it out after submission.

### DIFF
--- a/.github/workflows/user-input.yml
+++ b/.github/workflows/user-input.yml
@@ -1,0 +1,23 @@
+name: Manage awaiting user response
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  remove_label:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


hkt74 (commented):
it looks like there is an [existing action](https://github.com/google-gemini/generative-ai-python/blob/main/.github/workflows/remove-issue-labels.yml#L14-L18) to remove label, I found it in the old JS SDK

MarkDaoust (dismissed):

amirh:
> it looks like there is an [existing action](https://github.com/google-gemini/generative-ai-python/blob/main/.github/workflows/remove-issue-labels.yml#L14-L18) to remove label, I found it in the old JS SDK

Change to use this action 👍

MarkDaoust (approved):
Worth a shot.

PR #444, opened 2025-03-26
chore: Update user-input.yml GitHub workflow to trigger on PR comments

Annhiluc:


IvanLH (approved):

IvanLH:
Actually, do we need this? PRs have "request changes" state, which is what we are looking for no?

Annhiluc:
Yes, I think we can still use the "awaiting user response" tag here to align the triage process with open issues.

PR #461, opened 2025-04-12
Added an error message to nudge folks to the new SDK.

dynamicwebpaige:
Also updated the README. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,6 @@
-# Google AI SDK for JavaScript
+# Google Generative AI for Node.js
 
-> [!NOTE] A new Javascript/Typescript SDK, `@google/genai`
-> ([github](https://github.com/googleapis/js-genai/tree/main)), is currently
-> available in a *preview launch* - designed to work with Gemini 2.0 features
-> and support both the Gemini API and the Vertex API.
+> **IMPORTANT:** This SDK is intended for use with older Gemini models (e.g., Gemini 1.0 Pro, Gemini 1.5 Pro). For access to the latest models, including Gemini 2.0 and Gemini 2.5 Pro, please use the newer SDK available at [https://github.com/googleapis/js-genai](https://github.com/googleapis/js-genai).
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,9 +1,6 @@
-# Google AI SDK for JavaScript
+# Google Generative AI for Node.js
 
-> [!NOTE] A new Javascript/Typescript SDK, `@google/genai`
-> ([github](https://github.com/googleapis/js-genai/tree/main)), is currently
-> available in a *preview launch* - designed to work with Gemini 2.0 features
-> and support both the Gemini API and the Vertex API.
+> **IMPORTANT:** This SDK is intended for use with older Gemini models (e.g., Gemini 1.0 Pro, Gemini 1.5 Pro). For access to the latest models, including Gemini 2.0 and Gemini 2.5 Pro, please use the newer SDK available at [https://github.com/googleapis/js-genai](https://github.com/googleapis/js-genai).
```
rakyll, line 3: We can use a much stronger message similar to the Python package, see https://github.com/google-gemini/deprecated-generative-ai-python.

For older models, we want to encourage people to use the new SDK. This SDK will be around not to break the existing users. New comers to the old models should also use the new SDK ;)

pamorgan (approved):

pamorgan (changes requested):
Please make sure this is not a breaking change

src/requests/request.ts
```diff
@@ -196,46 +218,54 @@ async function handleResponseNotOk(
...
     }
+
   } catch (e) {
-    // ignored
+    // Error parsing JSON is ignored, proceed with original status/text
+  }
+
+  let finalMessage = `Error fetching from ${url.toString()}: [${response.status} ${response.statusText}] ${originalMessage || 'Server responded with an error.'}`; // Provide default if message is empty
+
+  if (isUnsupportedModelError) {
```
hkt74, line 258: this might be a breaking change if user has client side logic to process the `error message`. Maybe a safer approach is to log a warning message to the console?

